PR #8, opened 2026-02-20
Genre model and pytest config

Floko2005:
## Genre related models

This issue implements the 3 genre related models (_GenreUsAs_, _Genre_, _GenreTranslation_) in ```backend/genres/models.py```.

These models are the link between our database and the API, by creating and running the migrations, django will create the related tables.

The commands for generating and running the migrations are:

```bash
python manage.py makemigrations genres
python manage.py migrate
```
(**Note:** This can also be found in the README for future reference)

Make sure you run these commands while in the ```backend``` folder.

## Tests

For testing we are using pytest. The ```backend/pytest.ini``` file is pretty straightforward, it contains a reference to our settings and also defines which files it will run for testing.

If you make your own tests **make sure that the file follows the test_*.py naming convention**.

We require two packages for testing in the backend. So make sure if you test locally you have the following packages installed:

- pytest
- pytest-django

The folder structure for tests should be the same structure as found in ```backend/apps```.  For example, the genre models are defined in ```backend/apps/genres/models.py``` so our tests for these models should be implemented in the file ```backend/tests/genres/test_models.py```

This structure is nice to have, because every resource will have a few corresponding files:

- models.py for defining the models 
- serializer.py for defining  the json serializer
- admin.py for the admin page
- ...

With the current file structure, everything relating a resource can quickly be found together. 

## Discussion

I am unsure about using cascade deletion for the foreign keys. This is used in the Genre and GenreTranslation models:

```python
class Genre(BaseModel):
...
use_as = models.ForeignKey(
        GenreUseAs,
        on_delete=models.CASCADE,
```

```python
class GenreTranslation(BaseModel):
...
language = models.ForeignKey(
        Language,
        on_delete=models.CASCADE,
        ...
    )

    genre = models.ForeignKey(
        Genre,
        on_delete=models.CASCADE,
        ...
    )
...
```

I think it's useful because of someone deletes a language, I think it's logical that all the translations for that language are also deleted. The same counts that if a Genre is deleted, the translation for that genre can also be deleted.

The problem is that it is not very safe, accidents can happen and maybe somehow a language would accidentally be deleted which would cause a lot of data loss...

Do you think that it's ok to use Cascade for these 3?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,14 @@
 
 - django
 - djangorestframework
+- pytest
+- pytest-django
 
 ### Belangrijke commands:
 
-- start de dev server: python manage.py runserver
-- migrations aanmaken voor X: python manage.py makemigrations X
-    -- bv. python manage.py makemigrations languages
-- migrate: python manage.py migrate
+- start de dev server: ```python manage.py runserver```
+- migrations aanmaken voor X: ```python manage.py makemigrations X```
+    -- bv. ```python manage.py makemigrations languages```
+- migrate: ```python manage.py migrate```
 
 ---
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -3,3 +3,4 @@ venv
 
 *__pycache__
 *migrations
+.pytest_cache
--- a/backend/apps/genres/admin.py
+++ b/backend/apps/genres/admin.py
@@ -1,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
--- a/backend/apps/genres/models.py
+++ b/backend/apps/genres/models.py
@@ -1,0 +1,82 @@
+from django.db import models
+from apps.core.model import BaseModel
+from apps.languages.models import Language
+
+class GenreUseAs(BaseModel):
+    name = models.CharField(
+        max_length=50,
+        null=False,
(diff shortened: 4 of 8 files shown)


eliaryck:
I also understand the doubt - cascade can be risky.

However, in this case I think cascade is still appropriate. The relationships are clearly parent -> child, and the dependent records don’t make sense without their parent.

Since everything is managed through the admin, deletions are intentional - you have to explicitly click delete and confirm it. As long as only trusted users have delete permissions, the risk of accidental data loss is low.

So yes, i would use cascade for the three.

prince-b-b:
I think using _CASCADE_ for the deletions of GenreUseAs and Genre is appropriate. 

My first thought when it comes to the deletions of languages was to opt for _PROTECT_. Because, like you said, a lot of data could get erased unnecessarily if the user isn't being careful enough.
However, like Elias, I trust that the risk of accidental data loss is low because a limited set of users will have the ability to delete data. So, I agree that using _CASCADE_ is also fine in that scenario.

Floko2005:
This will be continued in PR #54 , Elias implemented the tests cleaner on a different branch which contains all changes on this branch. So this PR will be closed.

PR #67, opened 2026-02-21
Locations Data layer

### LINKED ISSUE #32: Locations Data Layer Implementation
## Objective 
Implement database models for the **Locations** domain and ensure data integrity through automated model tests.

## Scope
- Core and Translation models

## Technical Requirements
- Use the BaseModel

## Definition of Done

Floko2005:
## Location data layer implementation

This PR is related to issue #32

It implements the data layer for location related models:

- Location and LocationTranslation
- Space and SpaceTranslation
- Hall and HallTranslation

There is also an implementation for the factories and tests, but I am unsure if these will run (**see discussion below**)

There was also a small problem with a circular import between the Event and Production models, but this was easily fixed by removing the import in ```productions/models.py``` because it wasn't used.

## Discussion

- **CURRENTLY NO TEST IS ABLE TO RUN** until all models have been implemented. There are still a few undefined models left that have to be implementend, until that happens there will be errors in the code because there are a lot of relations. This means that I am still unsure if the tests will pass or fail. I still think that this should be merged with the backend branch, where we will (hopefully fast) have every model defined. After that we can make branches to fix failing tests. 

- Pytest uses caches, which doesn't allow us to use test files with the same file name (even if they are in a different folder). I renamed ```tests/genres/test_models.py``` to ```tests/genres/test_genre_models.py```. The tests for locations can be found in ```tests/locations/test_location_models.py```.


backend/apps/events/models.py
```diff
@@ -40,14 +42,20 @@ class Meta(BaseModel.Meta):
...
-                name="event_ends_after_starts",
-            )
-        ]
+        # TODO I get an error because of this, maybe it's better to use the clean method instead of a constraint for this?
+        #constraints = [
+        #    models.CheckConstraint(
+        #        check=Q(ends_at__gt=F("starts_at")),
+        #        name="event_ends_after_starts",
+        #    )
+        #]
```
eliaryck: It's because there is a newer method (check -> condition):
```suggestion
        constraints = [
            models.CheckConstraint(
                condition=Q(ends_at__gt=F("starts_at")),
                name="event_ends_after_starts",
            )
        ]
```

eliaryck (approved):
Very good for me, thank you! For the tests, we do indeed need to see if we have all the models and if they work.

prince-b-b (approved):
Looks good to me - I don't see anything inconsistent/incorrect regarding the model implementations or tests.

PR #69, opened 2026-02-21
Implemented a scraper for the viernulvier api.

### LINKED ISSUE #55: Scraper Viernulvier API
## Problem Statement

We need to integrate data from the Viernulvier API into our backend.  
Currently, there is no scraper/service responsible for fetching, transforming, and storing this external data.  

Without a structured scraping layer:
- External data integration becomes inconsistent
- Logic may end up duplicated across apps
- Future maintenance and scaling become harder

This issue defines how we implement and structure the scraper.

## Scope

**Included:**
- Implementing a scraper/service that fetches data from the [Viernulvier API](https://www.viernulvier.gent/api/docs#/)
- Transforming API responses into our internal models
- Error handling and basic logging
- Deciding on scraper architecture:
  - Per-app scraper (e.g. `apps/events/scraper.py`)
  - OR centralized scraper module (e.g. `backend/scrapers/viernulvier.py`)

**Excluded:**
- Frontend integration
- Advanced scheduling (Celery/cron) unless explicitly added later
- Full data sync strategy (incremental vs full sync) beyond basic implementation

## Acceptance Criteria
- [ ] A working scraper that can fetch and persist data from the [Viernulvier API](https://www.viernulvier.gent/api/docs#/)
- [ ] Clear folder structure decision (per-app vs centralized scraper module)
- [ ] Scraper logic separated from views and serializers
- [ ] Basic error handling for failed API requests
- [ ] Documentation added explaining where scraper logic lives

## Discussion
- Deciding on scraper architecture:
  - Per-app scraper (e.g. `apps/events/scraper.py`)
  - OR centralized scraper module (e.g. `backend/scrapers/viernulvier.py`)

tobitdsm:
Merging this pull request closes #55.

# Scrapers

The Viernulvier scraper lives in `apps/imports/scrapers/viernulvier.py` (centralized scraper module).

Use `sync_viernulvier(ModelClass, endpoint=...)` to fetch from `https://www.viernulvier.gent/api/v1` with `VIERNULVIER_API_KEY` and persist into the Django model.

To add a new scraper (useful in the future for e.g. UiTdatabank):
- create a new module in this folder
- expose a fetch_* and sync_* entry point
- keep network/transform/persist logic out of views/serializers

Example usage (e.g. in a management command):
```py
from apps.imports.scrapers.viernulvier import sync_viernulvier
from apps.events.models import ViernulvierItem

sync_viernulvier(ViernulvierItem, endpoint="/events")
```
You can also request specific parameters (server-side) like this:
```py
sync_viernulvier(ViernulvierItem, endpoint="/events", params={"created_at[after]": "2026-02-01T00:00:00+00:00"})
sync_viernulvier(ViernulvierItem, endpoint="/productions", params={"starts_at[strictly_before]": "2024-01-01T00:00:00+00:00"})
```
Preceding this with
```py
import logging
logging.basicConfig(level=logging.DEBUG)
```
will give you DEBUG logs when using `python manage.py shell`


tobitdsm:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/SELab-2/viernulvier-2/pull/69#pullrequestreview-3838347082)

tobitdsm:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/SELab-2/viernulvier-2/pull/69#pullrequestreview-3838425838)

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,135 @@
...
+
+def default_transform(item):
+    if not isinstance(item, dict):
+        raise ScraperError("Invalid item type for default_transform")
+    if "id" in item and item.get("id") is not None:
+        external_id = item.get("id")
+    elif "uuid" in item and item.get("uuid") is not None:
+        external_id = item.get("uuid")
+    else:
+        external_id = item.get("slug")
```
eliaryck: I think this one is a bit cleaner?
```suggestion
    
    external_id = (
        item.get("id")
        or item.get("uuid")
        or item.get("slug")
    )
```

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,135 @@
...
+    sync_viernulvier(Event, endpoint="/events")
+"""
+
+import logging
+import os
+from urllib.parse import urljoin, urlparse
+
+import requests
+from django.core.exceptions import FieldError
+from django.db import DatabaseError, IntegrityError
```
eliaryck: Needed for next suggestion
```suggestion
from django.db import DatabaseError, IntegrityError, transaction
```

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,135 @@
...
+        try:
+            model.objects.update_or_create(external_id=external_id, defaults=data)
+        except (IntegrityError, DatabaseError, FieldError) as exc:
+            logger.exception("Database error while syncing item: %s", external_id)
+            raise ScraperError("Database error while syncing item") from exc
+
+        saved += 1
+
+    logger.info("Synced %s items from Viernulvier", saved)
+    return saved
```
eliaryck: Implemented a robust, per-item atomic sync logic. This ensures a 'best effort' import where individual failures don't block the entire process, while maintaining data integrity. Your previous code blocked synchronization when an error occurred. But take another look at it :)
```suggestion
    items = fetch_viernulvier(endpoint=endpoint)
    
    saved = 0
    errors = 0
    seen = set()

    for item in items:
        external_id = None

        try:
            data = transform(item)

            if not isinstance(data, dict) or "payload" not in data:
                logger.error("Invalid transform output for item: %s", item)
                errors += 1
                continue

            external_id = data.pop("external_id", None)

            if not external_id:
                logger.warning("Skipping item without external_id: %s", item)
                errors += 1
                continue

            if external_id in seen:
                logger.warning("Duplicate external_id in batch: %s", external_id)
                continue

            seen.add(external_id)

            with transaction.atomic():
                model.objects.update_or_create(
                    external_id=external_id,
                    defaults=data,
                )

            saved += 1

        except (IntegrityError, DatabaseError, FieldError):
            logger.exception(
                "Database error while syncing item %s. Continuing.",
                external_id,
            )
            errors += 1
            continue

        except Exception:
            logger.exception(
                "Unexpected error while syncing item %s",
                item.get("id"),
            )
            errors += 1
            continue

    logger.info(
        "Viernulvier sync finished (%s). Saved=%s Errors=%s",
        endpoint,
        saved,
        errors,
    )

    return saved
```

eliaryck (changes requested):
Looks good, but I had one more important suggestion. Some tests will no longer work because of this, and you will have to adjust them. But very well done!

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _get_api_key():
+    api_key = os.getenv("VIERNULVIER_API_KEY")
+    if not api_key:
+        raise ScraperError("VIERNULVIER_API_KEY is not set")
+    return api_key
+
+
+def _build_headers():
+    return {"X-Api-Key": _get_api_key()}
```
eliaryck: The API of viernulvier expects this.
```suggestion
    return {
        "X-AUTH-TOKEN": api_key,
        "accept": "application/ld+json"
    }
```

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: As far as your implementation goes, not really if the id is '0'. However, I did clean up the code and wrote a simpler expression:
```py
external_id = None
    for key in ("id", "uuid", "slug"):
        if item.get(key) is not None:
            external_id = item.get(key)
            break
```

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: This might slow down big operations though, so I implemented this as a fallback when a batch operation might fail.

eliaryck (approved):
Great, I think this is good! But maybe do this one, because otherwise we'll get a 401 error. (https://github.com/SELab-2/viernulvier-2/pull/69/changes#r2838430247)

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: Right, good catch, missed that one.

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,175 @@
...
+def _get_api_key():
+    api_key = os.getenv("VIERNULVIER_API_KEY")
+    if not api_key:
+        raise ScraperError("VIERNULVIER_API_KEY is not set")
+    return api_key
+
+
+def _build_headers():
+    return {
+        "X-AUTH-TOKEN": api_key,
```
eliaryck: Sorry, I made a mistake, it should be: "X-AUTH-TOKEN": _get_api_key()

```suggestion
        "X-AUTH-TOKEN":  _get_api_key(),
```

eliaryck (approved):
Sorry for the mistake

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: Whoops, no worries!

backend/tests/scrapers/test_viernulvier.py
```diff
@@ -0,0 +1,689 @@
...
+
+from apps.imports.scrapers import viernulvier
+
+
+def test_fetch_uses_api_key_header(monkeypatch):
+    monkeypatch.setenv("VIERNULVIER_API_KEY", "test-key")
+
+    def fake_get(url, headers, timeout):
+        assert url.endswith("/events")
+        assert headers["X-Api-Key"] == "test-key"
```
Floko2005:
```suggestion
        assert headers["X-AUTH-TOKEN"] == "test-key"
```

You forgot to change this test after applying the suggestion from Elias.

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,175 @@
...
+        return data["items"]
+    if isinstance(data, dict) and "data" in data:
+        return data["data"]
+    if isinstance(data, list):
+        return data
+
+    raise ScraperError("Unexpected Viernulvier API payload")
+
+
+def default_transform(item):
```
Floko2005: Maybe add some documentation/explanation about what this does? I don't really understand what is happening here.

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,175 @@
...
+import os
+from urllib.parse import urljoin, urlparse
+
+import requests
+from django.core.exceptions import FieldError
+from django.db import DatabaseError, IntegrityError, transaction
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://viernulvier.gent/api"
```
Floko2005: They also use v1 in the URL:

Image

You could change the BASE_URL to reflect this:

```suggestion
BASE_URL = "https://viernulvier.gent/api/v1"
```

Or maybe put somewhere in documentation that the endpoints should always start with a version, in that case DEFAULT_ENDPOINT should be changes to ```"/api/events"```

Floko2005 (changes requested):
A lot of tests, which all pass (if you apply the one suggestion) so that is very good!

Did you test this with the actual implemented models? We currently have no external_id but this can easily be added in the BaseModel in ```apps/core/model.py```. But when I try to run this locally I just get errors, can you explain this a bit more how we can run the scraper so we can see actual data in our datbase?

After you have done this and the suggestions you can request a review of me again so I can approve this merge.

backend/tests/scrapers/test_viernulvier.py
tobitdsm: Yes, thanks!

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: This just extracts the id from the API, but I added a docstring now.

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: Right, I added the v1 to the base URL now.

tobitdsm:
@Floko2005
> Did you test this with the actual implemented models? We currently have no external_id but this can easily be added in the BaseModel in `apps/core/model.py`. But when I try to run this locally I just get errors, can you explain this a bit more how we can run the scraper so we can see actual data in our datbase?

The scraper sends a GET request from the requested endpoint to the server, and receives the data in JSON format. It then makes a dictionary with two fields:
- external_id: the id of the item for quick lookup
- payload: a JSON object with all fields that were received from the server. These should normally match the models (apart from the fields we discussed to leave out).

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,180 @@
...
+import os
+from urllib.parse import urljoin, urlparse
+
+import requests
+from django.core.exceptions import FieldError
+from django.db import DatabaseError, IntegrityError, transaction
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://viernulvier.gent/api/v1"
```
eliaryck: When you make a call, you get a 301, but with www. in front, this is no longer the case.
```suggestion
BASE_URL = "https://www.viernulvier.gent/api/v1"
```

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,180 @@
...
+    api_key = os.getenv("VIERNULVIER_API_KEY")
+    if not api_key:
+        raise ScraperError("VIERNULVIER_API_KEY is not set")
+    return api_key
+
+
+def _build_headers():
+    return {
+        "X-AUTH-TOKEN":  _get_api_key(),
+        "accept": "application/ld+json"
```
eliaryck: When I attempted to make an API call, I received an error. Your code expects JSON, hence this adjustment.
```suggestion
        "accept": "application/json"
```

eliaryck (changes requested):
Now I only get the error that it cannot connect to the database, so I think the scraper works with these changes :)

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: Okay, makes sense.

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: Yes indeed, I believe this got mixed up when applying requested changes, sorry!

eliaryck (dismissed):
Now I have no further comments ;) Perhaps add this somewhere in the wiki or explain it in a pull request. You will also see your logs when you test with: ```python manage.py shell```

```
import logging
logging.basicConfig(level=logging.DEBUG)

from apps.imports.scrapers.viernulvier import sync_viernulvier
from apps.events.models import Event

sync_viernulvier(Event, endpoint="/events")

**Review by Floko2005 — CHANGES_REQUESTED** [review]
I ran the postgres server locally on the correct port, I ran all migrations and could see the tables in the database. When I run the following in the manage.py shell:

```python
import logging
logging.basicConfig(level=logging.DEBUG)

from apps.imports.scrapers.viernulvier import sync_viernulvier
from apps.events.models import Event

sync_viernulvier(Event, endpoint="/events")
```
I get the following output: ```INFO:apps.imports.scrapers.viernulvier:Viernulvier sync finished: Saved=0, Errors=30```

And a lot of ```WARNING:apps.imports.scrapers.viernulvier:Skipping item without external_id: ...```

Does this work for you if you try to run the scraper, or do you get the same issue?

eliaryck:
> import logging
> logging.basicConfig(level=logging.DEBUG)
> 
> from apps.imports.scrapers.viernulvier import sync_viernulvier
> from apps.events.models import Event
> 
> sync_viernulvier(Event, endpoint="/events")

I think we should add in default_transform: ```external_id = next((item.get(key) for key in ("id", "uuid", "slug", "vendor_id") if item.get(key) is not None), None)``` So adding the vendor_id (I think that's unique for every item in their database.). And to have the external_id in each app, we should add the field in the ```class BaseModel(models.Model):```: ```external_id = models.CharField(max_length=255, unique=True)```. 

If I understand correctly, the external_id is used to check which items are already in the database and have therefore been fetched? 

I think with this changes it should work, but check it @tobitdsm. :) But maybe it will work if we have our views and serializers? I don't know if your scraper uses our models directly or not.

tobitdsm:
@Floko2005 
> And a lot of `WARNING:apps.imports.scrapers.viernulvier:Skipping item without external_id: ...`

Turns out the api doesn't return an id when requesting with json. I rewrote the scraper to now use ld+json, which includes an "\@id" field. Let's hope this works now...

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,597 @@
...
+        except FieldDoesNotExist:
+            # Try converting from camelCase to snake_case
+            snake_key = _camel_to_snake_case(key)
+            try:
+                field = model._meta.get_field(snake_key)
+            except FieldDoesNotExist:
+                # Field doesn't exist in model, skip it silently
+                logger.debug("Skipping unknown field '%s' for model %s", key, model.__name__)
+                continue
+
```
Floko2005, line 336: I had an issue when running the scraper:

```
File "/home/flo/Documents/Ugent/Jaar3/Sem2/Sel2/viernulvier-2/backend/apps/imports/scrapers/viernulvier.py", line 392, in _build_model_defaults
    if field.primary_key:
       ^^^^^^^^^^^^^^^^^`
```

So I asked copilot to take a look at it, and it suggested the following change:

```suggestion

        # Skip reverse relations (e.g., ManyToOneRel) that aren't real model fields
        if not isinstance(field, models.Field):
            logger.debug("Skipping reverse relation '%s' for model %s", key, model.__name__)
            continue

```

After applying that change, I can import items from the API! They appeared in my database.
```
INFO:apps.imports.scrapers.viernulvier:Viernulvier sync finished: Saved=30, Errors=0
```
 But I would like if you could double check if this suggestion doesn't break anything else.

There is also another issue the viernulvier API uses pages, and right now I think you only import the first page. I took a look at the API and saw that they use 30 items per page (that's why the import only saved 30 items).

But I think that this can easily be fixed! If you take a look at the respone from the API you can find a property 'view':

```json
"view": {
    "@id": "/api/v1/events?page=1",
    "@type": "PartialCollectionView",
    "first": "/api/v1/events?page=1",
    "last": "/api/v1/events?page=135",
    "next": "/api/v1/events?page=2"
  }
```

I think the easiest implementation would be to keep fetching the 'next' page, you can stop the loop when there isn't a next field anymore:

```json
"view": {
    "@id": "/api/v1/events?page=135",
    "@type": "PartialCollectionView",
    "first": "/api/v1/events?page=1",
    "last": "/api/v1/events?page=135",
    "previous": "/api/v1/events?page=134"
  }
```

I think you're very close to finishing the scraper, good job!

Floko2005 (changes requested):

Floko2005:
There is also a problem with the import_logs. After running the scraper I see no entries appear in the import_log table

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: The fix doesn't seem to break anything, and I now added pagination support.

tobitdsm:
@Floko2005 
> There is also a problem with the import_logs. After running the scraper I see no entries appear in the import_log table

Right, I completely forgot about that one. It seems that this was added after my PR was opened, so I don't have these files. I merged a new pull request (#106) to update my branch.

eliaryck:
I'll wait a bit before reviewing until you're done (maybe put it in draft mode), but it works from what I've tested! Some models are missing fields (but those are things we need to adjust in the models), but we'll solve those problems later.

eliaryck (approved):
Image 

Image
Yes, it works! 

We'll put the endpoints in a separate file, then add some logic and possibly adjust the models. But very good scraper and testing, great!

Floko2005 (changes requested):
There are still a few issues left:

- The import fails for Event and I can see an error appear in my console: ```year 0 is out of range```. I think that the code in parse_datetime generates this error, because if I put a try except around the function calls to ```parse_datetime``` the error doesn't occur.

- I see in the output after importing: ```INFO:apps.imports.scrapers.viernulvier:Viernulvier sync finished: Saved=4031, Errors=17``` but the problem is that I see no error message appear for these 17 in the import_log table. Maybe these 17 errors are related to the date issue, and the try except which I used to bypass these. You should check if this issue is resolved after fixing the date issue.

- At the moment only errors appear in the import_log, I think that we chose to put all imports in the table so this includes the successes.

tobitdsm:
I also updated the description of the PR to explain how parameter requests work.

eliaryck (approved):
This works for me! That parameter support is very useful and necessary. So then we have to create another file in every app where we scrape, I guess? I quickly generated something with AI, for example like this: (because, for example, theirs is external-order_link and ours is ticketing_url and then references to halls and such for events):
**apps/events/scrapers.py**
```python
"""
Event-specific Viernulvier scraper.

Transforms the raw API payload to match the Event model field names
before handing off to the generic sync mechanism.
"""

from typing import Any, Dict, List, Optional
from apps.imports.scrapers.viernulvier import fetch_viernulvier, sync_viernulvier, ScraperError
from apps.events.models import Event


def _pick_ticketing_url(external_order_url: Any) -> str:
    """Extract a single URL from the external_order_url dict.

    The API returns a dict like {"nl": "https://...", "fr": "https://..."}.
    We prefer Dutch, then fall back to the first available value.

    Args:
        external_order_url: Raw API value (dict, str, or None).

    Returns:
        A single URL string, or empty string if none found.
    """
    if not external_order_url:
        return ""
    if isinstance(external_order_url, str):
        return external_order_url
    if isinstance(external_order_url, dict):
        return (
            external_order_url.get("nl")
            or external_order_url.get("fr")
            or next(iter(external_order_url.values()), "")
        )
    return ""


def _transform_event_item(item: Dict[str, Any]) -> Dict[str, Any]:
    """Remap API fields to match the Event model field names.

    Args:
        item: Raw normalized API item.

    Returns:
        Transformed item ready for _build_model_defaults.
    """
    transformed = dict(item)

    # Map external_order_url → ticketing_url
    if "external_order_url" in transformed:
        transformed["ticketing_url"] = _pick_ticketing_url(
            transformed.pop("external_order_url")
        )

    return transformed


def sync_events(params: Optional[Dict[str, str]] = None) -> int:
    """Fetch and sync Viernulvier events into the Event model.

    Args:
        params: Optional query parameters, e.g. {"created_at[after]": "2024-01-01T00:00:00Z"}.

    Returns:
        Number of records created or updated.

    Raises:
        ScraperError: When fetching fails.
    """
    items = fetch_viernulvier(endpoint="/events", params=params)
    transformed = [_transform_event_item(item) for item in items]

    # Temporarily monkey-patch fetch so sync_viernulvier uses our transformed data.
    # Cleaner alternative: inline the sync logic (see note below).
    # Here we use a small helper that reuses the existing sync internals directly.
    from apps.imports.scrapers.viernulvier import (
        _extract_item_pk_value,
        _build_model_defaults,
        _missing_required_fields,
    )
    from django.db import transaction, DatabaseError, IntegrityError
    from django.core.exceptions import FieldError, ValidationError
    from apps.import_log.models import ImportLog
    from django.utils import timezone
    import logging

    logger = logging.getLogger(__name__)

    source = "viernulvier:/events (events app)"
    import_log = ImportLog.objects.create(
        source=source,
        status=ImportLog.Status.IN_PROGRESS,
        started_at=timezone.now(),
    )

    try:
        saved, errors, seen = 0, 0, set()

        with transaction.atomic():
            for item in transformed:
                if not isinstance(item, dict):
                    errors += 1
                    continue

                item_id = _extract_item_pk_value(Event, item)
                if item_id is None:
                    errors += 1
                    continue
                if item_id in seen:
                    continue
                seen.add(item_id)

                sid = transaction.savepoint()
                try:
                    defaults = _build_model_defaults(Event, item)
                    Event.objects.update_or_create(
                        **{Event._meta.pk.name: item_id},
                        defaults=defaults,
                    )
                    transaction.savepoint_commit(sid)
                    saved += 1
                except (IntegrityError, DatabaseError, FieldError, ValidationError):
                    transaction.savepoint_rollback(sid)
                    logger.error("DB error on item id=%s", item_id, exc_info=True)
                    errors += 1

        import_log.records_total = len(transformed)
        import_log.records_imported = saved
        import_log.records_failed = errors
        import_log.finished_at = timezone.now()
        import_log.status = (
            ImportLog.Status.SUCCESS if errors == 0
            else ImportLog.Status.PARTIAL_SUCCESS if saved > 0
            else ImportLog.Status.FAILED
        )
        import_log.save()
        return saved

    except Exception as exc:
        import_log.status = ImportLog.Status.FAILED
        import_log.finished_at = timezone.now()
        import_log.error_message = str(exc)
        import_log.save()
        raise
```

Then you call it up from, for example, a management command:
```
from apps.events.scrapers import sync_events

sync_events(params={"created_at[after]": "2024-01-01T00:00:00Z"})
```

Is this how you see it, or how would you go about doing those specific things for separate apps?

Floko2005 (changes requested):
There are still some problems when I tried scraping Events, you have done a really good job already but it would be nice if these last few problems could be fixed :)

1. There are still some issues with the error messages in the import_log:

```python
if errors == 0:
            import_log.status = ImportLog.Status.SUCCESS
        elif saved > 0:
            import_log.status = ImportLog.Status.PARTIAL_SUCCESS
        else:
            import_log.status = ImportLog.Status.FAILED
            import_log.error_message = f"All {errors} records failed to import"
```

- First of all, in a partial success there are errors as well so you should also include the error message in that case.
- Second of all the error message should be used to know what went wrong so this can be fixed. Currently the ```errors``` variable just holds the error count, which we can already find in the ```records_failed``` field, so it doesn't need to be in the message.
If the error comes from the viernulvier API you could probably forward that error message. If it comes from the scraper itself because of a parsing error or something than ofcourse it should be clear from the message where the error occured.
**I talked about this with Elias and he will put a suggestion to fix this.**

2. There are some issues with the related fields of the model that we are scraping. For example when I scrape the Events, I will find that the hall_id is null for every Event, ofcourse this is because no halls are scraped.

I think the best solution to this would be while you are scraping, if you come across a relation that has not been imported yet in our database, than you should first scrape the object in that relation. This should be done recursively, so for example I fetch an Event and I come across a Hall that has not been imported yet, then I should import that Hall. But if the Hall has relations that aren't imported yet, I should import those as well, and so on ...

We also don't want to work their ID's, this is because maybe there could be overlap if we import something from te csv's for example, or someone makes an item directly in the cms. So you should add the external_id field in ```apps/core/model.py```:

```python
import django.db.models as models


class BaseModel(models.Model):
    """Base class for all models in the project.

    Let all models inherit from this class and implement model validation
    in `clean()`. On `save()`, `full_clean()` is executed automatically.
    """

    external_id = models.PositiveIntegerField(
        null = True,
        blank = True,
    )
    
    class Meta:
        abstract = True
    
    def save(self, *args, **kwargs):
        """Always run validation before saving."""
        self.full_clean()
        super().save(*args, **kwargs)
```

**Don't forget to run the migrations! Otherwise tests will fail**
```python manage.py makemigrations```
```python manage.py migrate```

If you think that there are better ways to do this scraping of all related objects, feel free to suggest them! I am unsure if this is the best way, it's the only way I could think of.

File: backend/apps/imports/scrapers/viernulvier.py [hunk 12]
```diff
@@ -0,0 +1,734 @@
...
+Usage:
+    from apps.imports.scrapers.viernulvier import sync_viernulvier
+    from apps.events.models import Event
+
+    sync_viernulvier(Event, endpoint="/events")
+"""
+
+import logging
+import os
+import re
```
eliaryck, line 10:
```suggestion
import re
import sys
```

backend/apps/imports/scrapers/viernulvier.py
```diff
@@ -0,0 +1,734 @@
...
+        )
+        return saved
+
+    except Exception as exc:
+        # Update import log with error status
+        import_log.status = ImportLog.Status.FAILED
+        import_log.finished_at = timezone.now()
+        import_log.error_message = str(exc)
+        import_log.save()
+        raise
```
eliaryck, line 573: I have added error_messages list.

```suggestion
    try:
        items = fetch_viernulvier(endpoint=endpoint, params=params)

        saved = 0
        errors = 0
        error_messages = []
        seen = set()

        with transaction.atomic():
            for item in items:
                # Validate item is a dict
                if not isinstance(item, dict):
                    msg = f"Item is not a dict: {item}"
                    logger.error(msg)
                    errors += 1
                    error_messages.append(msg)
                    continue

                # Extract @id
                item_id = _extract_item_pk_value(model, item)

                # Validate item_id is not empty
                if item_id is None:
                    msg = f"Missing @id for item: {item}"
                    logger.warning(msg)
                    errors += 1
                    error_messages.append(msg)
                    continue

                # Check for duplicates in current batch
                if item_id in seen:
                    msg = f"Duplicate @id in batch: {item_id}"
                    logger.warning(msg)
                    error_messages.append(msg)
                    continue

                seen.add(item_id)

                # Persist item with savepoint isolation
                sid = transaction.savepoint()
                try:
                    defaults = _build_model_defaults(model, item)
                    missing = _missing_required_fields(model, defaults)
                    if missing:
                        msg = f"Missing fields for @id {item_id}: {', '.join(missing)}"
                        logger.warning(msg)
                        transaction.savepoint_rollback(sid)
                        errors += 1
                        error_messages.append(msg)
                        continue

                    obj, created = model.objects.update_or_create(
                        **{model._meta.pk.name: item_id},
                        defaults=defaults,
                    )
                    transaction.savepoint_commit(sid)
                    saved += 1
                except ValidationError as e:
                    transaction.savepoint_rollback(sid)
                    # Make ValidationErrors readable
                    messages = []
                    for field, errs in e.message_dict.items():
                        for err in errs:
                            if field == "__all__":
                                messages.append(err)
                            else:
                                messages.append(f"{field}: {err}")
                    msg = f"Validation error for @id {item_id}: {', '.join(messages)}"
                    logger.error(msg)
                    errors += 1
                    error_messages.append(msg)

                except (IntegrityError, DatabaseError, FieldError):
                    transaction.savepoint_rollback(sid)
                    exc_type, exc_value, exc_tb = sys.exc_info()
                    msg = f"Database error for @id {item_id}: {exc_type.__name__}: {exc_value}"
                    logger.error(msg, exc_info=True)
                    errors += 1
                    error_messages.append(msg)

        # Update import log with final status
        import_log.records_total = len(items)
        import_log.records_imported = saved
        import_log.records_failed = errors
        import_log.finished_at = timezone.now()

        if errors == 0:
            import_log.status = ImportLog.Status.SUCCESS
        elif saved > 0:
            import_log.status = ImportLog.Status.PARTIAL_SUCCESS
            import_log.error_message = f"{errors} records failed to import: {', '.join(error_messages)}"
        else:
            import_log.status = ImportLog.Status.FAILED
            import_log.error_message = f"All {errors} records failed to import: {', '.join(error_messages)}"

        import_log.save()

        logger.info(
            "Viernulvier sync finished: Saved=%s, Errors=%s",
            saved,
            errors,
        )
        return saved

    except Exception as exc:
        # Update import log with error status
        import_log.status = ImportLog.Status.FAILED
        import_log.finished_at = timezone.now()
        import_log.error_message = str(exc)
        import_log.save()
        raise
```

eliaryck (changes requested):
Now the error messages also work when partial is successful and display the errors nicely. (if you apply these suggestions)

tobitdsm:
@eliaryck 
> Is this how you see it, or how would you go about doing those specific things for separate apps?

This is a possible approach, but I feel like you're re-implementing a lot of logic that I already put in the scraper. It is indeed hard to match all the fields, especially the renamed ones, but I don't know if it may be possible to use an intermediate django model that only keeps the data we want in fields that match their api, and then convert that into the model we actually use? This would need to be implemented for each model, but if that isn't possible we will indeed need an approach similar to what you suggested.

backend/apps/imports/scrapers/viernulvier.py
tobitdsm: Nice, this is indeed better.

eliaryck:
Oops, sorry, I wanted to update that branch so I could start a new branch here to test certain things.

tobitdsm:
@Floko2005 
> I think the best solution to this would be while you are scraping, if you come across a relation that has not been imported yet in our database, than you should first scrape the object in that relation. This should be done recursively, so for example I fetch an Event and I come across a Hall that has not been imported yet, then I should import that Hall. But if the Hall has relations that aren't imported yet, I should import those as well, and so on ...

I implemented this (kind of) in [_resolve_fk_value](https://github.com/SELab-2/viernulvier-2/blob/c184eac31eacacb340b2f421a7d478ba965b1df6/backend/apps/imports/scrapers/viernulvier.py#L513C5-L513C22). This will now lookup the external id in the database, and try to add the related item when it can't find it (and fetch it from the api if necessary). Because this uses the same functions, this will happen recursively automatically.

eliaryck:
Can you just remove it in core/model.py and put that external ID in models.py? And delete the model.py file.

tobitdsm:
> Can you just remove it in core/model.py and put that external ID in models.py? And delete the model.py file.

Oops, my bad. I resolved the merge not knowing the file moved. Should be fixed now.

eliaryck (approved):
The tests work, but the coverage could be higher. Otherwise, well done, great job! Finally merge, hahah

Floko2005 (approved):
This is very nice! Good job, this will be very useful

PR #70, opened 2026-02-21
36 media library data layer implementation

### LINKED ISSUE #36: Media Library Data Layer Implementation
## Objective 
Implement database models for the **Media Library** domain and ensure data integrity through automated model tests.

## Scope
- Core and Translation models

## Technical Requirements
- Use the BaseModel

## Definition of Done

eliaryck:
Introduced a new `media_library` app with the following models and tests:

- `MediaGallery`
- `MediaItem`
- `MediaItemTranslation`
- `MediaItemCrop`

Features:
- Proper relationships (ForeignKeys with related names)
- Unique constraints where needed
- Ordering and string representations
- App registered in `settings`

## Integration

- Connected `MediaGallery` to the `productions` app via a `ForeignKey`.
- Added `media_library` to `INSTALLED_APPS`.

### DIFF
--- a/backend/apps/media_library/models.py
+++ b/backend/apps/media_library/models.py
@@ -1,0 +1,144 @@
+from django.db import models
+from apps.core.model import BaseModel
+from apps.languages.models import Language
+
+class MediaGallery(BaseModel):
+    """Model representing a media gallery."""
+    name = models.CharField(
+        max_length=255,
+        db_comment="Name of the media gallery.",
+    )
+
+    class Meta(BaseModel.Meta):
+        db_table = "media_gallery"
+        verbose_name = "Media Gallery"
+        verbose_name_plural = "Media Galleries"
+    
+    def __str__(self):
+        return self.name
+    
+class MediaItem(BaseModel):
+    """Model representing a media item."""
+    gallery = models.ForeignKey(
+        MediaGallery,
+        on_delete=models.CASCADE,
+        related_name="media_items",
+        db_comment="The media gallery this item belongs to.",
+    )
+
+    type = models.CharField(
+        max_length=50,
+        db_comment="Type of media item (e.g., image, video).",
+    )
+
+    format = models.CharField(
+        max_length=50,
+        blank=True,
+        db_comment="Format of the media item (e.g., jpg, mp4).",
+    )
+
+    original_filename = models.CharField(
+        max_length=255,
+        blank=True,
+        db_comment="Original filename of the media item.",
+    )
+
+    position = models.PositiveIntegerField(
+        default=0,
+        db_comment="Position of the media item within the gallery.",
+    )
+
+    width = models.PositiveIntegerField(
+        null=True,
+        blank=True,
+        db_comment="Width of the media item in pixels.",
+    )
+
+    height = models.PositiveIntegerField(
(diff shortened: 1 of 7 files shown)


Floko2005 (commented):
I am a bit confused with where we actually store the media. If we store it on our server, I think we need a ```path``` field in MediaItem, which holds the path to the actual file.

backend/apps/media_library/models.py
```diff
@@ -0,0 +1,144 @@
...
+class MediaItem(BaseModel):
+    """Model representing a media item."""
+    gallery = models.ForeignKey(
+        MediaGallery,
+        on_delete=models.CASCADE,
+        related_name="media_items",
+        db_comment="The media gallery this item belongs to.",
+    )
+
+    type = models.CharField(
```
prince-b-b, line 35: Is there a limited set of media item types? If so, I think that the `choices` attribute would come in handy:

```python
class MediaItemType(models.TextChoices):
    IMAGE = "image", "Image"
    VIDEO = "video", "Video"
    AUDIO = "audio", "Audio"


type = models.CharField(
    max_length=20,
    choices=MediaItemType.choices,
    db_comment="Type of media item."
)

**Comment by eliaryck** [comment]
> I am a bit confused with where we actually store the media. If we store it on our server, I think we need a `path` field in MediaItem, which holds the path to the actual file.

MediaItemTranslation has the URL path. And in MediaItemCrop, the URL then refers to our server somewhere.

image

File: backend/apps/media_library/models.py (hunk 1 above)
Comment (eliaryck) [reply backend/apps/media_library/models.py:35]: I changed it, that's a good suggestion.

File: backend/tests/media_library/test_media_library.py [hunk 2]
```diff
@@ -0,0 +1,198 @@
...
+# =====================================================
+
+class TestMediaItem:
+
+    def test_requires_gallery(self):
+        item = MediaItemFactory.build(gallery=None)
+        with pytest.raises(ValidationError):
+            item.full_clean()
+
+    def test_position_defaults_to_zero(self):
```
prince-b-b, line 64: In `MediaItem`, `position` has `default=0`:

```python
position = models.PositiveIntegerField(
    default=0,
    db_comment="Position of the media item within the gallery.",
)
```

However, `MediaItemFactory` always sets position using a `Sequence`:
```python
position = factory.Sequence(lambda n: n)
```
As a result, this test is misleading since this only works for the very first generated instance and does not actually test the model's default value.

My suggestion is to replace the test with one that directly tests the model default:
```python
def test_position_defaults_to_zero(self):
    gallery = MediaGalleryFactory()
    item = MediaItem.objects.create(gallery=gallery, type="image")
    assert item.position == 0
```

This way, you are testing the model's default behavior instead of the factory configuration.

Floko2005:
> > I am a bit confused with where we actually store the media. If we store it on our server, I think we need a `path` field in MediaItem, which holds the path to the actual file.
> 
> MediaItemTranslation has the URL path. And in MediaItemCrop, the URL then refers to our server somewhere.
> image

Alright, thanks for the clarification.

Floko2005 (approved):

prince-b-b (approved):

PR #74, opened 2026-02-22
44 imports data layer implementation

### LINKED ISSUE #44: Imports Data Layer Implementation
## Objective 
Implement database models for the **Imports** domain and ensure data integrity through automated model tests.

## Scope
- Core and Translation models

## Technical Requirements
- Use the BaseModel

## Definition of Done

Floko2005:
## Implementation of the ImportLog data layer

This PR closes issue #44

## Scope

This PR implemented the following:

- The model ImportLog which uses the 'enum like' TextChoises for status:
```python
class Status(models.TextChoices):
        PENDING = 'PENDING', 'Pending'
        IN_PROGRESS = 'IN_PROGRESS', 'In Progress'
        PARTIAL_SUCCESS = 'PARTIAL_SUCCESS', 'Partial Success'
        SUCCESS = 'SUCCESS', 'Success'
        FAILED = 'FAILED', 'Failed'
```
This is a lot simpler than defining a different table and referencing an entry with a foreign key. The downside is that with the enum you can't add a new Status externally,  you would need to add the new option by writing it in the code. But there aren't a lot of statuses we need here so this really is not a big problem. That's why I chose for this enum.  

-  ImportLogFactory which is used for testing
-  Some tests which all pass without warnings.


prince-b-b (approved):
The implementation of the `ImportLog` model looks great. I have one suggestion regarding a test in `test_import_log_models.py` - but since all tests work without any warnings, there's definitely no need to apply it.

```python
with pytest.raises(ValidationError):
    with transaction.atomic():
        ImportLogFactory(started_at=..., finished_at=...)
```

This test assumes a `ValidationError`, but `CheckConstraint` is enforced at the database level, which raises an `IntegrityError` unless you explicitly call `.full_clean()`.

To test the database constraint directly:
```python
from django.db import IntegrityError, transaction

def test_finished_must_be_after_started():
    started = timezone.now()
    finished = started - timedelta(minutes=5)

    with transaction.atomic():
        with pytest.raises(IntegrityError):
            ImportLogFactory(
                started_at=started,
                finished_at=finished,
            )
```

eliaryck (approved):
I think it's good, well done again!

PR #78, opened 2026-02-23
Issue 51: Add model tests for core

### LINKED ISSUE #51: Add model tests for Core
The **Core** model exists, but there are currently no automated tests for it. Without tests, changes to the model could introduce bugs or break functionality.

## Scope
- **Included:** Writing unit tests for **Core** using `pytest`, `Factory Boy`, and `Faker`.
- **Excluded:** Other models, views, or serializers.

## Acceptance Criteria
- [ ] Factory is created for **Core** using `Factory Boy`.
- [ ] Tests cover:
  - Field constraints (e.g., required fields, unique constraints)
  - Model methods and properties
  - Relationships (ForeignKeys, ManyToMany, etc.)
- [ ] Tests can run independently using an in-memory SQLite database.
- [ ] Tests are added to the app’s test suite and pass consistently.

## Discussion
Using factories and `pytest` ensures maintainable, readable, and fast tests. This makes future refactoring safer and ensures model behavior is verified automatically.

prince-b-b:
## Summary

- I implemented a factory for the `Core` model in `tests/factories/core.py` and worked out some tests in `tests/core/test_core_models.py` (all of them pass).
- I added the `core` to the local apps in the `settings` folder.

### DIFF
--- a/backend/config/settings/base.py
+++ b/backend/config/settings/base.py
@@ -51,6 +51,7 @@ LOCAL_APPS = [
     'apps.pricing',
     'apps.locations',
     'apps.media_library',
+    'apps.core'
     # TODO add more local apps here
 ]
 
--- a/backend/config/settings/base.py
+++ b/backend/config/settings/base.py
@@ -51,7 +51,6 @@ LOCAL_APPS = [
     'apps.pricing',
     'apps.locations',
     'apps.media_library',
-    'apps.core'
     # TODO add more local apps here
 ]
 
--- a/backend/tests/core/test_core_models.py
+++ b/backend/tests/core/test_core_models.py
@@ -1,3 +1,5 @@
+from unittest.mock import patch
+
 import pytest
 from django.core.exceptions import ValidationError
 from django.db import connection
--- a/backend/tests/core/test_core_models.py
+++ b/backend/tests/core/test_core_models.py
@@ -16,11 +16,6 @@ def core_dummy_table():
     with connection.schema_editor() as schema_editor:
         schema_editor.delete_model(CoreDummy)
 
-def drop_core_dummy_table():
-    with connection.schema_editor() as schema_editor:
-        schema_editor.delete_model(CoreDummy)
-
-
 class TestBaseModel:
     def test_basemodel_is_abstract(self):
         from apps.core.model import BaseModel
--- a/backend/tests/core/test_core_models.py
+++ b/backend/tests/core/test_core_models.py
@@ -8,7 +8,6 @@ from tests.factories.core import CoreDummy, CoreDummyFactory
 
 pytestmark = pytest.mark.django_db(transaction=True)
 
-
 @pytest.fixture(autouse=False)
 def core_dummy_table():
     """Create the CoreDummy table before each test and drop it afterward."""
--- a/backend/tests/core/test_core_models.py
+++ b/backend/tests/core/test_core_models.py
@@ -7,16 +7,14 @@ from tests.factories.core import CoreDummy, CoreDummyFactory
 pytestmark = pytest.mark.django_db(transaction=True)
 
 
-def create_core_dummy_table():
-    """
-    Creates the table for CoreDummy in the test database.
-
-    Because CoreDummy is a test-only model (not part of migrations),
-    its schema is created manually for these tests.
-    """
+@pytest.fixture(autouse=False)
+def core_dummy_table():
+    """Create the CoreDummy table before each test and drop it afterward."""
     with connection.schema_editor() as schema_editor:
         schema_editor.create_model(CoreDummy)
-
+    yield
+    with connection.schema_editor() as schema_editor:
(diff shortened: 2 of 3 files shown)


backend/config/settings/base.py
```diff
@@ -51,6 +51,7 @@
     'apps.pricing',
     'apps.locations',
     'apps.media_library',
+    'apps.core'
```
eliaryck: It's duplicate, and that is why we get an error when running the tests. I think you added that afterwards, which is why the tests ran on your computer. 
```suggestion
```

backend/tests/core/test_core_models.py
```diff
@@ -0,0 +1,47 @@
...
+def create_core_dummy_table():
+    """
+    Creates the table for CoreDummy in the test database.
+
+    Because CoreDummy is a test-only model (not part of migrations),
+    its schema is created manually for these tests.
+    """
+    with connection.schema_editor() as schema_editor:
+        schema_editor.create_model(CoreDummy)
+
```
eliaryck: Instead of manually calling create_core_dummy_table() in each test, it is better to use this fixture. Then the table will also drop automatically. :)
```suggestion
@pytest.fixture(autouse=False)
def core_dummy_table():
    """Create the CoreDummy table before each test and drop it afterward."""
    with connection.schema_editor() as schema_editor:
        schema_editor.create_model(CoreDummy)
    yield
    with connection.schema_editor() as schema_editor:
        schema_editor.delete_model(CoreDummy)
```

backend/tests/core/test_core_models.py
```diff
@@ -0,0 +1,47 @@
...
+
+def drop_core_dummy_table():
+    with connection.schema_editor() as schema_editor:
+        schema_editor.delete_model(CoreDummy)
+
```
eliaryck: This can be removed due to the above.
```suggestion
```

backend/tests/core/test_core_models.py
```diff
@@ -0,0 +1,47 @@
...
+            drop_core_dummy_table()
+
+    def test_save_persists_valid_instance(self):
+        create_core_dummy_table()
+        try:
+            obj = CoreDummyFactory.build(name="valid")
+            obj.save()
+            assert obj.pk is not None
+        finally:
+            drop_core_dummy_table()
```
eliaryck: The try and catch can be removed based on the above suggestion. And I would do some more testing to be really complete.

```suggestion
    def test_save_calls_full_clean(self, core_dummy_table):
        """save() must delegate to full_clean(), regardless of validity."""
        obj = CoreDummyFactory.build(name="ok")
        with patch.object(obj, "full_clean", wraps=obj.full_clean) as mock_clean:
            obj.save()
            mock_clean.assert_called_once()

    def test_save_persists_valid_instance(self, core_dummy_table):
        obj = CoreDummyFactory.build(name="valid")
        obj.save()
        assert obj.pk is not None

    def test_save_valid_instance_is_retrievable(self, core_dummy_table):
        obj = CoreDummyFactory.build(name="hello")
        obj.save()
        fetched = CoreDummy.objects.get(pk=obj.pk)
        assert fetched.name == "hello"

    def test_save_max_length_boundary_is_accepted(self, core_dummy_table):
        """A name of exactly 10 characters (the max_length) must be valid."""
        obj = CoreDummyFactory.build(name="a" * 10)
        obj.save()
        assert obj.pk is not None

    def test_save_blocks_name_too_long(self, core_dummy_table):
        """Names longer than max_length=10 must be rejected."""
        obj = CoreDummyFactory.build(name="this_is_way_too_long")
        with pytest.raises(ValidationError):
            obj.save()

    def test_save_blocks_blank_name(self, core_dummy_table):
        """blank=False means an empty string must be rejected."""
        obj = CoreDummyFactory.build(name="")
        with pytest.raises(ValidationError):
            obj.save()

    def test_save_blocks_null_name(self, core_dummy_table):
        """null=False means None must be rejected."""
        obj = CoreDummyFactory.build(name=None)
        with pytest.raises(ValidationError):
            obj.save()

    def test_save_on_update_also_runs_full_clean(self, core_dummy_table):
        """Validation must run not just on insert but also on update."""
        obj = CoreDummyFactory.build(name="valid")
        obj.save()

        obj.name = "this_is_way_too_long"
        with pytest.raises(ValidationError):
            obj.save()

    def test_save_valid_update_persists(self, core_dummy_table):
        obj = CoreDummyFactory.build(name="first")
        obj.save()
        obj.name = "second"
        obj.save()
        assert CoreDummy.objects.get(pk=obj.pk).name == "second"
```

eliaryck (changes requested):
I found a better solution for the try catch and I would do some more testing anyway (I suggested it here). :)

eliaryck (approved):
Now it's good :) You mist some suggestions to fix the tests. I have applied these :)

Floko2005 (approved):
Everything looks good, I see no issues with the tests.

PR #108, opened 2026-02-27
[DOC] Professionalize API Documentation (OpenAPI / ReDoc)

eliaryck:
# Documentation & API improvements

## OpenAPI / Swagger

- Enriched the schema metadata: added a proper API description, contact details, and license information to `SPECTACULAR_SETTINGS`.
- Added a **ReDoc** endpoint at `api/redoc/` alongside the existing Swagger UI at `api/docs/`, giving consumers a second documentation interface better suited for reading than testing.
- Added `schemas.py` to every app, following the pattern established in the `pricing` app. All `extend_schema` / `extend_schema_view` annotations are now kept out of `views.py` so view files stay focused on routing and queryset configuration only.

## Authentication

- Brought the `ApiKeyAuthentication` implementation into full compliance with [[RFC 7235](https://www.rfc-editor.org/rfc/rfc7235)](https://www.rfc-editor.org/rfc/rfc7235): the `WWW-Authenticate` header is now correctly included on `401 Unauthorized` responses, and the distinction between `401` (missing or invalid credentials) and `403` (valid credentials, insufficient scope) is applied consistently across the codebase and its tests.
- Updated the test suite to assert the correct status codes in line with the RFC — requests with no key now expect `401`, requests with a public key attempting a write now expect `403`.

## Documentation

- Added module-level docstrings, class docstrings, and inline field comments to `models.py`, `admin.py`, `serializers.py`, `views.py`, and `schemas.py` across all apps (`core`, `productions`, `events`, `tags`, `pricing`, `imports`).
- Every `help_text` on model fields is now mirrored in the serializer's `extra_kwargs` so drf-spectacular picks them up and renders them in both Swagger UI and ReDoc without duplication in the schema decorators.

## All tests pass and admin still works normally. Not much has been changed apart from the documentation (an extra field here and there in the models, perhaps).


backend/apps/events/models.py
```diff
@@ -1,100 +1,193 @@
+"""
+Models for the Events app.
+
+Events are scheduled occurrences of productions:
+
+    Event  ->  EventPrice
```
Floko2005: I think you mean Production here instead of EventPrice?

```suggestion
    Event  ->  Production
```

If not, then I think the wording is a bit confusing because when the previous line talks about the relationship between events and productions and ends with a ':' then It feels like the next line should still say something about this relation.

backend/apps/genres/models.py
```diff
@@ -53,53 +74,46 @@ class Meta(BaseModel.Meta):
...
-        # Get the Dutch name for the genre
-        dutch_name = GenreTranslation.objects.filter(
-            genre=self,
-            language_id="nl"
-        ).first()
-
-        # Show both translations in the string representation of the genre
-        return f"{self.type} - [{str(english_name)}] - [{str(dutch_name)}]"
+    def __str__(self) -> str:
+        return self.type
```
Floko2005, line 78: I think we should use a base language to always put in the string, (not only for this, but for multiple models that have their name in different languages).

The reason that I think we should do this is because in the admin panel, if we want to add for example a genre to production I think normally someone would start typing the name of the genre for finding it and  add it to a production. If we only print the type, we can't easily find the exact genre we want. Plus it's possible for multiple genres to use the same type, so we would see multiple entries with the same string representation.

Image

Floko2005: I don't think that this should be fixed on this branch, I just wanted to let you know. I think it's best if we create an issue to fix this problem later?

Floko2005 (changes requested):
This is amazing, everything is very well documented. I took my time reading through everything and apart from my comments everything is really well done!

backend/apps/genres/models.py
eliaryck: I have created an issue for this. (#116)

prince-b-b (approved):
Elias, kudos for the effort you put in this - the documentation is sufficiently thorough and should give first-time readers an understandable and clear view of the different aspects of the written code. :-)

Floko2005 (approved):

PR #117, opened 2026-02-28
114 api throttling

eliaryck:
## Summary

This PR introduces rate limiting for the shared `PUBLIC_API_KEY` using Django REST Framework’s `SimpleRateThrottle`.

Because the public key is shared by all consumers, throttling is applied **per client**, not per key. Clients are differentiated using a fingerprint composed of:

- Client IP address (via DRF `get_ident()`)
- `User-Agent` header

The combination is hashed (SHA-256) before being stored in the cache to avoid persisting raw IPs or headers.

Two limits are enforced:

- Per-minute limit (`PublicKeyMinuteThrottle`)
- Per-hour limit (`PublicKeyHourThrottle`)

The `INTERNAL_API_KEY` remains unrestricted via a no-op `InternalKeyThrottle`.

## Environment Configuration

- **Production** -> Throttling enabled  
- **Test** -> Throttling enabled (required for throttle-related tests)  
- **Development** -> Throttling disabled to avoid friction during local development  

This ensures production protection without impacting developer workflow.

image

*If you want to test it, you need to set those rates in `dev.py` and then use curl via the terminal, for example.*


Floko2005 (changes requested):
Everything works like it should, very nice!

Floko2005 (approved):

tobitdsm (approved):
Nice, good idea. I have no clue if the rates you configured will strike the balance we want, but we can change those later if we like.

PR #138, opened 2026-03-04
Use base language for display strings in admin and api

prince-b-b:
_Related to issue #116_ 

## Summary

This pull request implements the requirements of the issue by ensuring that base-language display strings are consistently used across the admin interface and API responses.

Changes included:

- Added base-language `display_*` fields (e.g. `display_name`, `display_title`, `display_description`) to serializers where applicable, while preserving full translation dictionaries.
- Updated model `__str__` methods so Django admin uses the base-language representation (with sensible fallbacks to avoid duplicate labels like `X (X)`).
- Updated existing tests that validate serializer field exposure to account for the newly introduced `display_*` fields.
- Added targeted serializer tests to verify that `display_*` fields:
  - use the base language when available (derived from `settings.LANGUAGE_CODE`), and
  - fall back to the first available translation when the base language translation is missing.

## Notes

- Run the full test suite to ensure no regressions were introduced.
- Verify that the end goal of the issue is met:
  - Admin uses base-language strings via `__str__`
  - API exposes base-language display strings via `display_*` fields while keeping full translation data.

### DIFF
--- a/backend/apps/core/models.py
+++ b/backend/apps/core/models.py
@@ -8,6 +8,7 @@ a record is written to the database.
 """
 
 from django.db import models
+from django.conf import settings
 
 
 class BaseModel(models.Model):
@@ -75,3 +76,37 @@ class BaseModel(models.Model):
         """
         self.full_clean()
         super().save(*args, **kwargs)
+
+    @classmethod
+    def base_language_code(cls) -> str:
+        """
+        Returns the project's base language code.
+        """
+        code = getattr(settings, "LANGUAGE_CODE", "en")
+        return code.split("-")[0].lower()
+
+    def get_base_translation(self, related_name="translations"):
+        """
+        Returns the translation object for the base language,
+        falling back to the first available translation.
+        """
+        manager = getattr(self, related_name, None)
+        if not manager:
+            return None
+
+        base_code = self.base_language_code()
+        return (
+            manager.filter(language__code=base_code).first()
+            or manager.first()
+        )
+
+    def get_base_display_name(
+        self,
+        related_name="translations",
+        name_field="name",
+        fallback=None,
+    ):
+        tr = self.get_base_translation(related_name=related_name)
+        if not tr:
+            return fallback
+        return getattr(tr, name_field, fallback)
--- a/backend/apps/core/models.py
+++ b/backend/apps/core/models.py
@@ -95,10 +95,8 @@ class BaseModel(models.Model):
             return None
 
         base_code = self.base_language_code()
-        return (
-            manager.filter(language__code=base_code).first()
-            or manager.first()
-        )
+        qs = manager.all()
+        return qs.filter(language__code=base_code).first() or qs.first()
 
     def get_base_display_name(
         self,
@@ -109,4 +107,4 @@ class BaseModel(models.Model):
         tr = self.get_base_translation(related_name=related_name)
         if not tr:
             return fallback
-        return getattr(tr, name_field, fallback)
+        return getattr(tr, name_field, None) or fallback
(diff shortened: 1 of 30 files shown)


backend/apps/core/models.py
```diff
@@ -74,4 +75,38 @@ def save(self, *args, **kwargs) -> None:
...
+        """
+        manager = getattr(self, related_name, None)
+        if not manager:
+            return None
+
+        base_code = self.base_language_code()
+        return (
+            manager.filter(language__code=base_code).first()
+            or manager.first()
+        )
```
eliaryck: We get the N+1 problem so:

```suggestion
        qs = manager.all()
        return qs.filter(language__code=base_code).first() or qs.first()
```

backend/apps/core/models.py
```diff
@@ -74,4 +75,38 @@ def save(self, *args, **kwargs) -> None:
...
+    def get_base_display_name(
+        self,
+        related_name="translations",
+        name_field="name",
+        fallback=None,
+    ):
+        tr = self.get_base_translation(related_name=related_name)
+        if not tr:
+            return fallback
+        return getattr(tr, name_field, fallback)
```
eliaryck: I think we'll have problems with the empty string otherwise.
```suggestion
        return getattr(tr, name_field, None) or fallback
```

backend/apps/core/serializers.py
```diff
@@ -104,4 +105,122 @@ def get_translated_field(
...
+                    return value
+
+        # Fallback: first non-empty value
+        for t in translations:
+            value = getattr(t, field_name, None)
+            if value:
+                return value
+
+        # Final fallback
+        return fallback
```
eliaryck:
```suggestion
        first_available = None

        for t in translations:
            value = getattr(t, field_name, None)
            if not value:
                continue
            
            # If this translation matches the base language, return it immediately
            if t.language.code == base_code:
                return value
            
            # Otherwise, keep track of the first available translation as a fallback
            if first_available is None:
                first_available = value

        return first_available or fallback
```

backend/apps/events/models.py
```diff
@@ -116,7 +116,9 @@ def clean(self) -> None:
             raise ValidationError("Event end time must be after start time.")
 
     def __str__(self) -> str:
-        return f"Event {self.id} - {self.production} @ {self.starts_at}"
+        production = str(self.production)
```
eliaryck: To avoid problems when self.production is unknown.
```suggestion
        production = str(self.production) if self.production else "Unknown Production"
```

eliaryck (changes requested):
Well done, I have a few more suggestions.

eliaryck (approved):
Well done! All the tests pass also.

Floko2005 (changes requested):
Everything looks very good, this naming is really a lot better. But I think you forgot to do this renaming for tags :)

```apps/tags/models.py```
```python
def __str__(self) -> str:
        name = self.get_base_display_name(
            related_name="translations",
            name_field="name",
            fallback=None,
        )
        return name or f"Tag {self.id}"
``` 

I think it should look something like this?

If you implement it that way, you should change the ```test_str_representation``` in ```tests/tags/test_tag_models.py```

If you change it to this the test should pass again:

```python
def test_str_representation(self):
        tag = TagFactory(type="genre")
        TagTranslationFactory(tag=tag, language__code="en", name="Rock")
        assert str(tag) == "Rock"
```

I want to add these as suggestions but I don't know how to do that to files that aren't listed in the current changes, sorry!

Floko2005 (approved):
Great! This can be merged :)

eliaryck (approved):
Good!

PR #234, opened 2026-03-11
api client setup

Floko2005:
This PR does the initial setup for using an API client to use for contacting our backend.

To test this you should make sure that you have all the required packages for frontend and backend:

- Frontend: ```npm install``` should do the trick
- Backend: ```pip install -r requirements/base.txt```

Make sure that both are running with the commands:

- Frontend: ```npm run dev```
- Backend: ```python manage.py runserver```

Now you can test if this by temporarily changing the home page in ```src/pages/HomePage.tsx``` to:

```tsx
import { Container, Paper, Stack, Typography } from '@mui/material'
import { useEffect } from 'react'
import { useTranslation } from 'react-i18next'
import { api } from '../services/api'

const HomePage = () => {
  const { t } = useTranslation()

  useEffect( () => {
    const fetchData = async() => {
      try {
        const data = await api.get("/events/");
        console.log(data.data);
      } catch (error) {
        console.error("Error fetching data:", error);
      }
    }

    fetchData();
  }, []);

  return (
    <Container maxWidth="md" sx={{ py: 6 }}>
      <Paper elevation={3} sx={{ p: 4 }}>
        <Stack spacing={3}>
          <Typography variant="h3" component="h1">
            {t('title')}
          </Typography>
          <Typography variant="subtitle1">{t('subtitle')}</Typography>
        </Stack>
      </Paper>
    </Container>
  )
}

export default HomePage
```

If you now open your network tab and you refresh your page, you should see a succesful request that goes to the /events endpoint of our API.

If there are any problems please let me know!


frontend/src/services/api.ts
```diff
@@ -0,0 +1,25 @@
+import axios from 'axios'
+
+const API_URL: string = import.meta.env.VITE_API_URL
```
JDLMRN:
```suggestion
```

frontend/src/services/api.ts
```diff
@@ -0,0 +1,25 @@
...
+ *   const res = await api.get("/endpoint/");
+ *   console.log(res.data); // Do something with the response data
+ * } catch (error) {
+ *   console.error("Error fetching data:", error);
+ * }
+ * ```
+ *
+ **/
+export const api = axios.create({
+  baseURL: API_URL,
```
JDLMRN:
```suggestion
  baseURL: '/api',
```

frontend/src/vite-env.d.ts
```diff
@@ -1 +1,10 @@
 /// <reference types="vite/client" />
+
+interface ImportMetaEnv {
+  readonly VITE_API_URL: string
```
JDLMRN:
```suggestion
```

frontend/.env.dev.example
```diff
@@ -0,0 +1,2 @@
+VITE_API_URL=http://127.0.0.1:8000/api
```
JDLMRN:
```suggestion
```

JDLMRN (changes requested):
You can remove the need for the `API_URL` environment variable since you don't need to add the domain in a fetch when the domain of the frontend and backend are the same, you can just use `/api` instead.

The rest looks fine to me! Good work!

xxxxnoahboyxxxx (approved):
It works for me, just don't forget Jasper's change requests

Floko2005:
> You can remove the need for the `API_URL` environment variable since you don't need to add the domain in a fetch when the domain of the frontend and backend are the same, you can just use `/api` instead.

This was a really good suggestion, thanks!

JDLMRN (approved):
Perfect, works as intended!

eliaryck (approved):

PR #260, opened 2026-03-16
256 api json response changes

### LINKED ISSUE #256: API JSON response change4s
## Problem Statement
Currently the API response gives only the id for related objects. This is inconvenient for the frontend, because now it requires to make multiple requests to the API for fetching all the related data. We can prevent this by including all the related data in one request.

## Scope
This issue occurs for the following models:

- MediaItem (gallery)
- Hall (space)
- Space (location)
- Genre (use_as)
- Event (production, hall)
- EventPrice (price, price_rank)

## Acceptance Criteria
- [ ] Where required, the related data is also sent in the request
- [ ] All tests pass

Floko2005:
# API json response changes

## Description
This PR solves the problem in issue #256

I am not 100% sure if this was the best fix for this problem, so please if you think of something better make sure to mention it! The biggest problem was that infinite recursion was possible, for example hall -> space -> halls -> hall -> space -> ...
To solve this I used new serializers specifically to prevent this recursion, if there is a better way just let me know and I will implement it like that.

I also added an option when retrieving a single production, to also get all the related events. For example: 
```.../api/productions/1?include=events```

## Related Issues
- closes #256 

## Type of Change
- [ ] Feature
- [ ] Bugfix
- [x] Refactor
- [ ] Documentation

## Testing Instructions
You should run the backend and frontend tests!
You can also test manually in the frontend if you add a useEffect in the home page and log some of the responses.
You can check the /api/docs as well to see if I forgot anything important, I hope not :)
I also changed the documentation in the schemas.py a bit, let me know if I missed something!

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests

### CODE REVIEW

**Review by prince-b-b — CHANGES_REQUESTED** [review]
I find this to be a good solution! The only thing I would change is the name of the class `NestedRepresentationPKField` so that it's more conventional, like `NestedFKSerializer` for example.

**Review by eliaryck — CHANGES_REQUESTED** [review]
```python
class SpaceSerializer(serializers.ModelSerializer, TranslatableSerializerMixin):
    """
    Represents a Space.

    The `name` field contains all available translations as a dictionary,
    for example: {"en": "Stage A", "fr": "Scène A"}.
    """

    name = serializers.SerializerMethodField(
        help_text=(
            "Dictionary containing all available translations of the space name "
            '(e.g. {"en": "Stage A", "fr": "Scène A"}). '
            "Read-only - use the translation endpoints to manage translations."
        ),
    )

    display_name = serializers.SerializerMethodField(help_text="Human-readable space name in the project's base language.")
    location = LocationSerializer(read_only=True)

    location_id = serializers.PrimaryKeyRelatedField(
        queryset=Location.objects.all(),
        source="location",
        write_only=True,
        help_text="ID of the parent **Location** this space belongs to.",
    )

    halls = serializers.SerializerMethodField(
        help_text="Halls that belong to this space, each including translated fields.",
    )

    class Meta:
        model = Space
        fields = ["id", "location", "location_id", "name", "display_name", "halls"]
        read_only_fields = ["id", "name", "display_name", "halls"]
        extra_kwargs = {
            "location": {},
        }
```
Can you take a look at this? It's a lot easier I think :)

backend/apps/media_library/serializers.py
```diff
@@ -138,11 +155,9 @@ class Meta:
             "credits",
             "link",
             "crops",
+            "gallery",
```
eliaryck: That’s already stated above.
```suggestion
```

backend/tests/events/test_event_views.py
```diff
@@ -471,8 +473,12 @@ def setUp(self):
 
     def test_list_prefetches_related_models(self):
         # Ensure query count stays bounded when related data grows
-        with self.assertNumQueries(6):
+        with CaptureQueriesContext(connection) as captured:
             response = self.client.get("/api/events/?ordering=id", **pub_headers())
 
+        for i, query in enumerate(captured.captured_queries):
+            print(f"{i + 1}: {query['sql'][:200]}")
```
eliaryck: I think this route might work?
```suggestion
```

eliaryck (changes requested):
Otherwise, it looks good to me—well done!

backend/tests/events/test_event_views.py
Floko2005: This print was just to check what wasn't being prefetched correctly, it can be removed now.

eliaryck (approved):
Well done!

prince-b-b (approved):
Changes look good! Tested locally and everything works as expected.

PR #277, opened 2026-03-21
Add MediaGalleryItem model and many-to-many relationship

### LINKED ISSUE #275: [BUG] The links between galleries and items are always null
## Problem Statement
Currently, the media_items always contain “null” as the ID of the corresponding gallery. As a result, the media_items in the gallery's response are always empty. The problem is that the links for galleries and items are visible in their API when you fetch galleries. So we need to fix it there somewhere.

## Scope
Backend/Scraper

## Steps to Reproduce
1. Fetch the galleries and look at the response
2. Fetch the items and look at the response

## Expected Behavior
Normally, those links should be there.

## Actual Behavior
The links doesn't exist.

## Screenshots

Image

Image

## Acceptance Criteria

## Discussion
So you might need to create a new table to handle the links properly. The problem is that the links for galleries and items are visible in their API when you fetch galleries. So we need to fix it there somewhere.

tobitdsm:
## Description
This PR solves a bug where the field `gallery_id` in the `media_item` table was always null. The fix is implemented in a similar way as the extra crops command.

## Related Issues
- closes #275 

## Type of Change
- [ ] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation

## Testing Instructions
1. Import all media galleries: `python manage.py sync_viernulvier --only media_galleries`
2. Import all media items: `python manage.py sync_viernulvier --only media_items`
3. Verify null fields by querying the database: `SELECT gallery_id FROM media_item;`
4. Import the gallery-item links: `python manage.py sync_viernulvier --only media_item_gallery_links`
5. Verify non-null fields by querying the database: `SELECT gallery_id FROM media_item;`

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests


eliaryck (changes requested):
You could also check to see if, when retrieving the crops, you can take the additional flags into account (e.g., created-after or something like that). That doesn't work yet.

eliaryck (changes requested):

eliaryck (changes requested):
Image

eliaryck (changes requested):
Everything looks good; the test percentage could be just a little higher (if I hadn't mentioned it, Florian would have said something about it ;)

eliaryck (approved):

Floko2005 (approved):
Very nice, the db now has the gallery_id's correctly filled. The tests look good and coverage is on 100%. Good job

PR #288, opened 2026-03-23
added footer component (#162)

### LINKED ISSUE #162: [FE] Footer component
## Problem Statement
We need a footer component.

## Scope
Which page(s)/component(s) are affected?
Footer component
All pages

## Proposed Change
What exactly will be improved? Abstracting away the footer component.

## Acceptance Criteria
- [ ] UI improvement implemented as described
- [ ] No layout regressions on common breakpoints
- [ ] Accessibility improvement verified (where applicable)
- [ ] Reviewed and merged via PR

## Discussion
Screenshots, design references, notes.

ArneMaegherman:
## Summary
Closes #162

Implements the Footer component, abstracting it away from individual pages.
Includes NL/EN translations via react-i18next and a simplified nav structure.

## Changes
- Add `Footer.jsx` with 3-column layout (address, navigation, social)
- Add footer translations to existing `locales/nl/translation.json` and `locales/en/translation.json`
- Navigation links: Home, Series, Events

## Testing
- [ ] Verified layout on common breakpoints (desktop, tablet, mobile)
- [ ] Verified NL ↔ EN language switch
- [ ] No layout regressions on pages using Footer
- [ ] All tests still PASS


JDLMRN (changes requested):
I have a couple of remarks:
- The social icons and the link to the newsletter don't go anywhere
- The font size of the address is smaller than the other text in the footer
- When there is no space for all three columns to be next to eachother, I would just place them all below eachother (instead of splitting them 2-1)
- When the blocks are under eachother, I would center them on the page
- I would make all of the lines in the blocks to never wrap; on some widths the LinkedIn icon gets wrapped to a new line, and the same for the address

xxxxnoahboyxxxx (changes requested):
It is already starting to look good, but I agree with JDLMRN, so I'll look again after you make some changes. Though I would recommend looking if removing the page links from the footer when screen gets to small looks better.

xxxxnoahboyxxxx (approved):
- All links work now
- It looks much better now on smaller screens, but you might want to look into fixing this small problem just before the switch to a smaller screen:
- Font sizes look equal

Image

JDLMRN (approved):

JDLMRN:
> * All links work now
> * It looks much better now on smaller screens, but you might want to look into fixing this small problem just before the switch to a smaller screen:
> * Font sizes look equal
> 
> Image

I've fixed this issue in the latest commit, now there is always a set gap between the blocks.

xxxxnoahboyxxxx (approved):

PR #304, opened 2026-03-30
181: fe Updating the tag chip component

xxxxnoahboyxxxx:
# Description

In this PR, the GenreChip and Tag component have been combined into 1 GenreAndTagChip component. See #181 for more information.

I replaced all current GenreChip and Tag components with the new component, but it is not yet used correctly everywhere. So people who use one of these will need to make sure to update it. But maybe wait until the backend is updated for this.

## Related issues
- #181 
- #160
- #246 must be merged into dev first. 

## Type of Change
- [x] Feature
- [ ] Bugfix
- [x] Refactor
- [ ] Documentation
- [ ] Other (please specify): _____________

## Testing instructions

1. run npm run test

2. Replace pages\HomePage.tsx with

```ts
import { Box, Container, Paper, Stack, Typography, useMediaQuery, useTheme } from '@mui/material'
import { useEffect, useMemo, useState } from 'react'
import { useTranslation } from 'react-i18next'
import GenreAndTagChip from '../components/chips/GenreAndTagChip'
import SearchControlsBar from '../components/searchbar/SearchControlsBar'
import { useSearchBarUrlState } from '../components/searchbar/useSearchBarUrlState'
import type { Genre } from '../types/Genres'
import { getTranslatedRecord } from '../utils/translations'

// TODO: use the floatingAlerts when needed

const MOCK_SERIES_TAGS = [
  { id: 1, name: 'Familie', labels: {} },
  { id: 2, name: 'Jong publiek', labels: {} },
  { id: 3, name: 'Premiere', labels: {} },
]

const HomePage = () => {
  const { t, i18n } = useTranslation()
  const theme = useTheme()
  const isMobile = useMediaQuery(theme.breakpoints.down('md'))
  const [genres, setGenres] = useState<Genre[]>([])
  const {
    searchValue,
    sortTarget,
    sortDirection,
    viewMode,
    selectedGenreIds,
    selectedSeriesTagIds,
    setSearchValue,
    setSortTarget,
    setSortDirection,
    setViewMode,
    toggleGenreId,
    toggleSeriesTagId,
  } = useSearchBarUrlState({ isMobile })

  useEffect(() => {
    const loadGenres = async () => {
      try {
        const { getGenres } = await import('../services/genres/Genres')
        const response = await getGenres({ pageSize: 20 })
        setGenres(response.results)
      } catch {
        setGenres([])
      }
    }

    loadGenres()
  }, [])

  const genreChips = useMemo(
    () =>
      genres
        .map((genre) => ({
[...]

PR #326, opened 2026-04-05
316 related productions component for productiondetailspage

### LINKED ISSUE #316: [FE][FEATURE] Related productions component for ProductionDetailsPage
## Problem Statement
This is needed so we can see all related productions ( ~productions that share a tag) on a ProductionDetailsPage like shown in the mockup.

## Scope
This only includes the component itself, and adding it to the page. 

## Acceptance Criteria
- [ ]  all tests should pass
- [ ]  it looks good, and fits in our current style
- [ ]  clicking on a listed production takes you to that production's detail page

## Design/Technical Approach
Add RelatedProductions.tsx component in ```components/productions```, implement this so it can be added in the ProductionDetailsPage.tsx

Floko2005:
# Related productions component

## Description
This component implements a generic carousel component which is used for the RelatedProductions component (and will be used for the media list which needs improvement). This component is used on the ProductionDetailsPage.

## Related Issues
- closes #316

## Type of Change
- [X] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Documentation

## Testing Instructions
You can run the tests which should all pass.
To test manually: 
1. go to the cms and create a new tag (VideoDroom)
2. go to the productions and search for videodroom, you can bulk add the newly created tag to these productions.
3. Now you can go to one of the productions holding the tag by going to ```/productions/:id``` in the frontend.
4. You should see a list of related productions at the bottom of the list

**Note:** I think testing with the db of the server will give issues, it didn't work for me. I think this is because maybe the most recent migrations haven't been executed.

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests


xxxxnoahboyxxxx:
@Floko2005  
I can't get the productions that I added a tag to to load.
I keep getting: Kon productie niet laden
and this from the local backend:
psycopg2.errors.UndefinedTable: relation "production_tag_translation" does not exist

You might want to ask someone else to review this. Unless this is not a problem on my end.

Floko2005:
> @Floko2005 I can't get the productions that I added a tag to to load. I keep getting: Kon productie niet laden and this from the local backend: psycopg2.errors.UndefinedTable: relation "production_tag_translation" does not exist
> 
> You might want to ask someone else to review this. Unless this is not a problem on my end.

Did you run the migrations? If you are not sure you can check in the cms:
image
If you see the Tag Translations your db should be setup correctly, if not you should try to run the migrations again.

xxxxnoahboyxxxx:
Yes, I always do this when stating the local backend:
```cmd
python manage.py makemigrations                              
python manage.py migrate
python manage.py runserver
```

And I even synced the productions.
I do have the tag translation in the cms. But when I try to open a production that I added a tag to in the frontend, I get an error:
```cmd
...
A lot of incomprehensible Traceback errors

django.db.utils.ProgrammingError: relation "production_tag_translation" does not exist
LINE 1: ..., "production_tag_translation"."description" FROM "productio...
                                                             ^

ERROR 2026-04-06 10:30:11,299 log 34712 42156 Internal Server Error: /api/v1/productions/1/
ERROR 2026-04-06 10:30:11,300 log 34712 42156 "GET /api/v1/productions/1/?include=events,related HTTP/1.1" 500 168
```

frontend/src/pages/ProductionDetailPage.tsx
```diff
@@ -69,11 +70,17 @@ const ProductionDetailsPage = () => {
...
 
   const [prod, setProd] = useState<Production | null>(null)
   const [loading, setLoading] = useState<boolean>(false)
   const [error, setError] = useState<string | null>(null)
 
+  // Keep a mutable reference to the latest translator so the data fetch effect remains id-driven.
+  useEffect(() => {
+    translatorRef.current = t
+  }, [t])
+
```
xxxxnoahboyxxxx: What is the advantage of doing the translation like this instead of just using t()?

xxxxnoahboyxxxx (changes requested):
- The arrow of the carousel is not perfectly in the middle. This irritates me ;D.
- Maybe see if the fallback picture can be changed so the `viernulvier` is in the middle.
- If I am reading the code correctly, the arrows are only supposed to appear when you hover over the component? But they are always there for me; it only disappears when the cursor leaves the page/screen.
- When there are a lot of productions, there are a lot of circles on small screens, but I have no idea how we could do this differently, so I would just leave it as is.
- The number of circles also only updates after you have pressed the arrow or a circle after you scale the screen. Can they be auto-updated when the screen changes size?

Image

frontend/src/pages/ProductionDetailPage.tsx
Floko2005: This is something I forgot to remove, initially I fetched the related productions separately from the initial production fetch. With that implementation I had the issue that changing the language would refetch the related productions, and that part fixed it.

But now I get the initial production and it's related productions in 1 initial fetch, so it isn't required anymore. I will remove it in one of the next commits :)

Floko2005: I was wrong! Apparently the useRef is required for some tests to pass. Whithout the use ref these failed:

ProductionDetailPage › shows load failed if getProduction throws and navigates to home
ProductionDetailPage › shows invalid id error and navigates to home for non-numeric ID

So I put the useRef back.

xxxxnoahboyxxxx (changes requested):
- Using numbering after a certain number of circles is better. But maybe make it so you can manually change the number so you don't need to keep clicking on the arrow.

- I see you added a link to the series:
 
Image

It's better to use the tag from #304, which already includes routing. Also, I see that the tag already gets added at the top of the page, though not yet correctly.

Image

It might be handy to have 1 link at the top and one at the bottom.
If there are multiple series tags, it might be better to render them all above the related carousel like you now do with the link.
Something to discuss as a group maybe?

Also are we just going to add all the related productions from the different tags into 1 carousel or a carousel per tag?

JDLMRN (changes requested):
Overall looks good! But I do have a couple of remarks.

1. The `Carousel` looks a bit overcomplicated. I don't think you need all of the resize listeners. Embla has a `slidesToScroll: 'auto'` option that automatically adjusts the slides for you based on the width of the container. Also changing the controls solely based on width can still give issues when for example 1000+ productions are added to a tag. So maybe the solution could be is to only show the amount of dots that fit on the screens width, and center the active dot. Then it will also work the same way on mobile as it does on desktop.
2. You can use the `ProductionGridCard` in the carousel.
3. The style of the tag label doesn't really fit in. I would either remove the blue color, or as Noah suggested put the chip there. Image
4. While you're at it, maybe change the media carousel above it to use the same newly added `Carousel` component?

Floko2005:
> Overall looks good! But I do have a couple of remarks.
> 
>     1. The `Carousel` looks a bit overcomplicated. I don't think you need all of the resize listeners. Embla has a `slidesToScroll: 'auto'` option that automatically adjusts the slides for you based on the width of the container. Also changing the controls solely based on width can still give issues when for example 1000+ productions are added to a tag. So maybe the solution could be is to only show the amount of dots that fit on the screens width, and center the active dot. Then it will also work the same way on mobile as it does on desktop.
> 
>     2. You can use the `ProductionGridCard` in the carousel.
> 
>     3. The style of the tag label doesn't really fit in. I would either remove the blue color, or as Noah suggested put the chip there. Image
> 
>     4. While you're at it, maybe change the media carousel above it to use the same newly added `Carousel` component?

I will look into it! Thanks for the feedback.
For nr. 4, I wanted to improve the media list in another PR, but it probably won't be too much code so I'll do it in this PR too.

Floko2005:
> * Using numbering after a certain number of circles is better. But maybe make it so you can manually change the number so you don't need to keep clicking on the arrow.
> 
>     * I see you added a link to the series:
> 
> 
> Image
> 
> It's better to use the tag from #304, which already includes routing. Also, I see that the tag already gets added at the top of the page, though not yet correctly.
> Image
> 
> It might be handy to have 1 link at the top and one at the bottom. If there are multiple series tags, it might be better to render them all above the related carousel like you now do with the link. Something to discuss as a group maybe?
> 
> Also are we just going to add all the related productions from the different tags into 1 carousel or a carousel per tag?

Could you tell me what is wrong with the way I add the tag at the top of the page?

About your question, the code now will make a carousel with related productions per carousel. The tag name (which links to the series page) will then be just above the corresponding carousel.

xxxxnoahboyxxxx:
> Could you tell me what is wrong with the way I add the tag at the top of the page?

I don't think this chip was added by you, but by an older PR.
It is still the older tag component, I am guessing. The updated one has not been merged yet. The current chip at the top of the page does not open the series page correctly, like your link. But this should be fixed once the new chip has been merged and used.

image

> About your question, the code now will make a carousel with related productions per carousel. The tag name (which links to the series page) will then be just above the corresponding carousel.

I would change the links with the GenreAndTagChip once that has been merged.

- Maybe make the font of the section title bigger, and see if it looks better in Bold text.

image

Floko2005:
@JDLMRN @xxxxnoahboyxxxx  could you take a look at it now? I have a few questions left before I want to implement this further.

First of all I want to ask if this way of showing the dots was what you meant? Any feedback is appreciated!

I started using the tag chips and it does look better, but the chip is permanently purple but I guess these will be fixed in PR #304? Once that PR is merged I will definitely take a look again to use the updated component.

I now use the ProductionGridCard component but have a few issues with it:

1. The cards can be of different size, which I think looks a bit weird in the carousel.
2. There were some overlap issues which I fixed by adding some extra padding in the carousel but now you can always see half of the first production card from the next tab which looks a bit weird, but I can't seem to fix this I will look into it tomorrow but I don't have time this evening. I think this could be fixed with some adjustments in the ProductionGridCard component.
3. The productionGridCard component requires an input of type Production, but the API returns related productions with minimal data only required for showing the title, artists and image. So it isn't possible to use these without build errors in the productionGridCard. At the moment I just fill the data with empty values so it is a complete production type. But I think it would be best to either change the API to return complete productions, or make the ProductionGridCard component work with the minimal data. What do you think would be the best option be?

So my main question is if I am allowed to adjust some things in the ProductionGridCard? It isn't the component I worked on so maybe you like it better if I stay away from it (which I would understand). I would add some optional parameters to fix some of these issues, without changing the core way it's used so no changes need to be made in other places where the gridcard is used.

I mainly think of the following change:

- Add an option that long titles and artists get cut, followed by '...'  so that the cards stay the same height. By hovering over the card you would be able to see the complete title and artist
- Make the font size more customizable
- Maybe have more control over the width of the component

xxxxnoahboyxxxx:
I think something might have gone wrong indeed xp.

image

I think the dots look ok, maybe use a lighter gray in dark mode (after #342 you should have a bunch of gray options). I would just look into whether you can make it so the number version can be manually adjusted instead of having to use the arrow.


> I started using the tag chips and it does look better, but the chip is permanently but I guess these will be fixed in PR https://github.com/SELab-2/viernulvier-2/pull/304?

Permanently what?

Floko2005:
> I think something might have gone wrong indeed xp.
> image
> 
> I think the dots look ok, maybe use a lighter gray in dark mode (after #342 you should have a bunch of gray options). I would just look into whether you can make it so the number version can be manually adjusted instead of having to use the arrow.
> 
> > I started using the tag chips and it does look better, but the chip is permanently but I guess these will be fixed in PR #304?
> 
> Permanently what?

I meant to say permanently purple, normally the tag chips aren't colored yet until you hover over them.
I also notice that your tag chip doesn't show the name? I don't know if you didn't name the tag when you created it? If you did you should see the tag's name instead of the ID. It's also weird that you still have the overlap... On my system I don't see any, even if I change the viewport:

image

About the numbers, I tried to implement Jasper's idea which makes the mobile and desktop view work the same;

>So maybe the solution could be is to only show the amount of dots that fit on the screens width, and center the active dot. Then it will also work the same way on mobile as it does on desktop.

I don't think that setting a number manually  would be very useful, I think the only specific use case is that someone would reference an image on a certain page number, but then there is the problem that different viewports can have different numbers so I don't think it would work very well. 

You don't have to press the arrows constantly at the moment, you can swipe (also available in desktop mode by holding and swiping with the mouse) which allows you to go past multiple tabs at once. It still isn't as fast as typing in the tab number you want of course.

If you really want this option I could implement it, but maybe you could give me some use cases for it to convince me :)

xxxxnoahboyxxxx:
> I meant to say permanently purple, normally the tag chips aren't colored yet until you hover over them. I also notice that your tag chip doesn't show the name? I don't know if you didn't name the tag when you created it? If you did you should see the tag's name instead of the ID. It's also weird that you still have the overlap... On my system I don't see any, even if I change the viewport:

It is purple because it is a series tag, i did that to differentiate between a genre and a tag. I had named it test so that is indeed weird. But That should normally be fixed when the new tag gets used. There is no point in trying to fix it now with the old version. It might be the fallback because i only made 1 translation.

> image
> About the numbers, I tried to implement Jasper's idea which makes the mobile and desktop view work the same;
> 
> > So maybe the solution could be is to only show the amount of dots that fit on the screens width, and center the active dot. Then it will also work the same way on mobile as it does on desktop.
> 
> I don't think that setting a number manually would be very useful, I think the only specific use case is that someone would reference an image on a certain page number, but then there is the problem that different viewports can have different numbers so I don't think it would work very well.
> 
> You don't have to press the arrows constantly at the moment, you can swipe (also available in desktop mode by holding and swiping with the mouse) which allows you to go past multiple tabs at once. It still isn't as fast as typing in the tab number you want of course.
> 
> If you really want this option I could implement it, but maybe you could give me some use cases for it to convince me :)

No, if you can do that, it is fine. It was mostly so you could go through the cards faster.

Floko2005:
> It is purple because it is a series tag, i did that to differentiate between a genre and a tag. I had named it test so that is indeed weird. But That should normally be fixed when the new tag gets used. There is no point in trying to fix it now with the old version. It might be the fallback because i only made 1 translation.
> 
Thanks for the clarification! I'll wait until the new version is merged and then I will take a look at it again.

frontend/src/components/carousel/Carousel.tsx
```diff
@@ -0,0 +1,303 @@
...
+    if (!el || !container) return
+
+    // Calculate the scroll position so that the dot is centered in the container
+    // Doing it this way won't force a scroll towards the caroussel on a refresh
+    const elLeft = el.offsetLeft
+    const elWidth = el.offsetWidth
+    const containerWidth = container.offsetWidth
+    const targetScroll = elLeft - containerWidth / 2 + elWidth / 2
+
+    container.scrollTo({ left: targetScroll, behavior: 'smooth' })
```
JDLMRN:
```suggestion
    let startTime: number | null = null

    function animateScroll(currentTime: number) {
      if (!startTime) {
        startTime = currentTime
      }

      const progress = Math.min((currentTime - startTime) / 650, 1)
      const ease = progress < 0.5 ? 2 * progress * progress : -1 + (4 - 2 * progress) * progress
      container!.scrollLeft = container!.scrollLeft + (targetScroll - container!.scrollLeft) * ease

      if (progress < 1) {
        requestAnimationFrame(animateScroll)
      }
    }
    requestAnimationFrame(animateScroll)
```
I would add a slight animation to the scroll on the dots, it makes it feel smoother.

frontend/src/components/carousel/Carousel.tsx
```diff
@@ -0,0 +1,303 @@
...
+            onPointerUp={() => (document.activeElement as HTMLElement | null)?.blur()}
+            disabled={!canScrollPrev && !loop}
+            className="carousel-nav"
+            sx={(theme) => ({
+              position: 'absolute',
+              top: '50%',
+              left: 0,
+              transform: 'translate(-50%, -50%)',
+              zIndex: 2,
+              border: `1px solid ${theme.palette.divider}`,
```
JDLMRN:
```suggestion
      {showArrows && hasMultipleSlides ? (
```
I would add this to remove the navigational arrows whenever there is only one slide.

frontend/src/components/production/MediaList.tsx
```diff
@@ -1,15 +1,13 @@
-import { useMediaQuery, useTheme } from '@mui/material'
-import useEmblaCarousel from 'embla-carousel-react'
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import CloseRoundedIcon from '@mui/icons-material/CloseRounded'
+import { Box, IconButton, Modal, useTheme } from '@mui/material'
+import { useState, type KeyboardEvent } from 'react'
+import Carousel from '../carousel/Carousel'
```
JDLMRN:
```suggestion

```

frontend/src/components/production/MediaList.tsx
```diff
@@ -1,15 +1,13 @@
...
 import type { MediaItem } from '../../types/Media'
```
JDLMRN, line 6:
```suggestion
import Carousel from '../carousel/Carousel'

```

frontend/src/components/production/MediaList.tsx
```diff
@@ -121,109 +87,139 @@ export default function MediaList({ mediaItems }: MediaListProps) {
...
+        onClose={closePreview}
+        slotProps={{
+          backdrop: {
+            sx: {
+              backgroundColor: 'rgba(35, 35, 35, 0.72)',
+            },
+          },
+        }}
+      >
+        <Box
```
JDLMRN, line 170:
```suggestion
        <Box
          onClick={closePreview}
```
I would add this (in combination with the change below this) to close the popup whenever you click outside of the image too.

frontend/src/components/production/MediaList.tsx
```diff
@@ -121,109 +87,139 @@ export default function MediaList({ mediaItems }: MediaListProps) {
...
-        style={{
-          marginTop: '12px',
-          textAlign: 'center',
-          color: theme.palette.text.secondary,
-          fontSize: '0.9rem',
-        }}
-      >
-        {`${displayIndex} / ${total}`}
-      </div>
+          <Box
```
JDLMRN, line 180:
```suggestion
          <Box
            onClick={(event) => event.stopPropagation()}
```
I would add this (in combination with the change above this) to close the popup whenever you click outside of the image too.

JDLMRN (changes requested):
Looking good! I've added some tiny suggestions, and if those are applied, it's approved by me. 

Very good job!

(you'll also need to fix a failing test in the `MediaList.test.tsx` after applying the changes; remove line 172 and remove the `fireEvent` import at the top)

JDLMRN (approved):
Well done!

frontend/src/components/production/RelatedProductions.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+        return (
+          <Box key={entry.tag.id}>
+            <Box sx={{ mb: 1.5, display: 'inline-flex', fontWeight: 600 }}>
+              <GenreAndTagChip
+                name={String(entry.tag.id)}
+                labels={entry.tag.name as Record<string, string>}
+                context="series"
+                id={entry.tag.id}
+              />
```
xxxxnoahboyxxxx, line 68:
```suggestion
              <GenreAndTagChip
                name={String(entry.tag.id)}
                labels={entry.tag.name as Record<string, string>}
                context="series"
                chipType="seriesTag"
                id={entry.tag.id}
              />
```

You need to mention the chiptype

frontend/src/components/production/RelatedProductions.tsx
```diff
@@ -0,0 +1,128 @@
...
+        {t('productions.detail.related', 'Related productions')}
+      </Typography>
+      {visibleResults.map((entry) => {
+        const tagName = getTagLabel(entry.tag)
+
+        return (
+          <Box key={entry.tag.id}>
+            <Box sx={{ mb: 1.5, display: 'inline-flex', fontWeight: 600 }}>
+              <GenreAndTagChip
+                name={String(entry.tag.id)}
```
xxxxnoahboyxxxx: You need to give Display_name as name (this will be the fallback if there is no translation)

Now this happens:

Image

xxxxnoahboyxxxx (changes requested):
Is it normal that cards without subtitle are smaller?

Image

frontend/src/pages/ProductionDetailPage.tsx
eliaryck:
```suggestion
  const mediaItems = production.media_gallery?.media_items
  const firstPhoto = mediaItems?.find((item) => item.type === 'foto')
```

eliaryck:
```suggestion
      {production.media_gallery?.media_items?.length > 0 && (
```

eliaryck (changes requested):

Floko2005:
> Is it normal that cards without subtitle are smaller?
> Image

When I look at the other places where the card is used, this does not happen. I'll fix it!

xxxxnoahboyxxxx (approved):
Looks good

eliaryck (approved):
Nice done!

PR #350, opened 2026-04-08
Filter panel for productions

### LINKED ISSUE #348: [BUG] Genre/tag chips are cut off in Production Cards
## Problem Statement
The genre/tag chips are being cut off in the `ProductionListCard` and `ProductionGridCard` components.

## Scope
Frontend:
- `ProductionListCard.tsx`
- `ProductionGridCard.tsx`

## Steps to Reproduce
- Go to the homepage

## Expected Behavior
The chips should be fully visible

## Actual Behavior
The bottom of the chips are cut off

## Screenshots

Image
Image

## Acceptance Criteria

### LINKED ISSUE #349: [FE][FEATURE] Filter panel for productions
## Problem Statement
We need a way to apply filters to search for certain productions.

## Scope
Frontend, create a dedicated component that can be later (re)used for the series as well.

## Acceptance Criteria
- [ ] UI: The UI needs to be clear
- [ ] Filtering: The filters need to actually work and output productions based on certain criteria
- [ ] Testing: There need to be tests to verify the correctness of the component

JDLMRN:
## Description
This PR adds a filter panel component to filter productions based on certain criteria that are defined in the backend.

## Related Issues
- Closes #349
- Closes #348

## Type of Change
- [x] Feature
- [ ] Bugfix
- [ ] Refactor
- [ ] Documentation

### Automated

Run the frontend unit tests and lint in your usual way (CI or local).

### Manual

Go to the homepage and you should see the filter panel on the left side of your screen. When changing certain filters, make sure the productions change accordingly.

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests


xxxxnoahboyxxxx (changes requested):
- Can you make it so that the years to choose from only go until the current year?

Image

- If you select both, it does nothing. If both can not be true at the same time, maybe make it so clicking one disables the other.

Image

- Same with attendance
- Something is very wrong with the selected tag style ;D.

Image

- Did you make it so there is only 1 chip per line? Because sometimes 2 could fit on 1 line.
- Only the first clicked genre reloads the cards
- Only deselecting all genres reloads the cards
- The static chips in the cards do not turn purple after you click them in the filter panel
- Are these supposed to be the SeriesTags? Because then you pull the wrong tags.

Image

JDLMRN:
> * Can you make it so that the years to choose from only go until the current year?

Good suggestion! I've added the `disableFuture` prop (native to the library) which disables dates in the future. It doesn't hide them, but this seems the easiest option.

> * If you select both, it does nothing. If both can not be true at the same time, maybe make it so clicking one disables the other.
> * Same with attendance

Good suggestion! Implemented.

> * Something is very wrong with the selected tag style ;D.

Hmm interesting, I just used your `GenreAndTagChip` without changing anything to it, so I think the problem lies within that component?

> * Did you make it so there is only 1 chip per line? Because sometimes 2 could fit on 1 line.

Yes, I chose that to make it easier to find genres/tags since they are sorted alphabetically.

> * Only the first clicked genre reloads the cards

Correct. I'm waiting for the backend to update to allow for multiple genres/tags.

> * Only deselecting all genres reloads the cards

This is because of the same issue talked about above. This will be fixed whenever the backend gets updated.

> * The static chips in the cards do not turn purple after you click them in the filter panel

Good suggestion! Implemented.

> * Are these supposed to be the SeriesTags? Because then you pull the wrong tags.

It currently fetches the tags from the `api/tags` endpoint which is as far as I know going to be the endpoint where the series will be made on? So I assume after the backend changes that will be fixed too?

xxxxnoahboyxxxx:
> > * Something is very wrong with the selected tag style ;D.
> 
> Hmm interesting, I just used your `GenreAndTagChip` without changing anything to it, so I think the problem lies within that component?

Probably when I changed the styling

> > * The static chips in the cards do not turn purple after you click them in the filter panel
> 
> Good suggestion! Implemented.

This was already supposed to be implemented if the correct chips are used, and a list of selected tags is kept.

> > * Are these supposed to be the SeriesTags? Because then you pull the wrong tags.
> 
> It currently fetches the tags from the `api/tags` endpoint which is as far as I know going to be the endpoint where the series will be made on? So I assume after the backend changes that will be fixed too?

I just found it weird that I could not find the series tags I created locally.

JDLMRN:
> Probably when I changed the styling

Probably. In my latest commit I've made it look like it was before.

> This was already supposed to be implemented if the correct chips are used, and a list of selected tags is kept.

Correct, I just had to pass a prop and that was it. You implemented it well!

> I just found it weird that I could not find the series tags I created locally.

So apparently the `/api/tags` is an unused endpoint, or not up to date yet. I've now altered it by using the same endpoint used in the `/series` page. If in the future the backend gets updated, we need to change it here as well.

frontend/src/components/chips/GenreAndTagChip.tsx
```diff
@@ -105,19 +98,11 @@ const GenreAndTagChip = ({
...
                 alignItems: 'center',
                 justifyContent: 'center',
-                width: 2.25,
-                height: 2.25,
-                borderRadius: '50%',
-                backgroundColor: hovered ? iconHoverBackground : 'transparent',
-                color: hovered ? selectedChipColor : 'inherit',
-                boxShadow: hovered ? `0 0 0 2px ${iconHoverBackground}` : undefined,
-                cursor: 'pointer',
-                transition: 'background 0.15s, color 0.15s',
```
xxxxnoahboyxxxx, line 115: Why did you remove the circle around the X? The circle was just made too small for some reason.

xxxxnoahboyxxxx (approved):
Looks good.

frontend/src/components/searchbar/useSearchBarUrlState.ts
```diff
@@ -302,14 +420,34 @@ export const useSearchBarUrlState = ({
...
+    setSearchParams(
+      (currentParams) => {
+        const nextParams = new URLSearchParams(currentParams)
+        nextParams.delete('eid')
+        return nextParams
+      },
+      { replace: true },
+    )
+  }, [searchParams, setSearchParams])
+
```
ArneMaegherman: What is the function of this code, `eid` never gets used as far as I can see.

ArneMaegherman (changes requested):
Looks good and works how it should (except for the parts that still need backend updates ofc). There are still a few linting warnings tho. 

- Would it be nice if the checkboxes became colored in the accent color (purple) when selected instead of grey, or not?
- I think the code in `useSearchBarUrlState.ts` could be made simpler now that AttendenceMode and PerformerType are a toggle. Could be wrong.
- And I don't like how you've called the folder for your components `filter-panel` instead of `filterpanel` or `filterPanel` but thats personal preference.

frontend/src/components/chips/GenreAndTagChip.tsx
JDLMRN: Since it looked so weird, I didn't that there was a circle around the X before. But actually I don't mind it like this? It's more subtle.

frontend/src/components/searchbar/useSearchBarUrlState.ts
JDLMRN: Good catch. It's dead code that I implemented for filtering productions based on an event ID but that wasn't needed in the end.

JDLMRN:
> Looks good and works how it should (except for the parts that still need backend updates ofc). There are still a few linting warnings tho.

Fixed the eslint warnings

> * Would it be nice if the checkboxes became colored in the accent color (purple) when selected instead of grey, or not?

I've changed them to use the secondary color and they look fine to me

> * I think the code in `useSearchBarUrlState.ts` could be made simpler now that AttendenceMode and PerformerType are a toggle. Could be wrong.

Yes indeed, fixed.

> * And I don't like how you've called the folder for your components `filter-panel` instead of `filterpanel` or `filterPanel` but thats personal preference.

I think it's fine for folders. It fits better with the all lowercase folder names for single word folders

ArneMaegherman (approved):
Good changes, good job.

PR #370, opened 2026-04-11
Fix tags translation

### LINKED ISSUE #368: [BUG] Tags/Genres translation doesn't work
## Problem Statement
There is a bug in the tag localization system. When a user switches the application language (e.g., from Dutch to English), certain tags fail to translate. For example, the tag **"at the vooruit"** remains in English even when the rest of the interface has been switched to Dutch.

## Scope
* **Frontend:** Tag components and display labels.
* **API:** The JSON payload returning tag data (checking if the backend provides translated strings or raw keys).
* **Database/CMS:** Verification that translation entries exist for these specific tags.

## Steps to Reproduce
1. Log in to the application and set the language to **English**.
2. Navigate to an item or list containing the tag "at the vooruit".
3. Change the language settings to **Dutch**.
4. Observe the tag label in the list or detail view.

## Expected Behavior
The tag should dynamically update to match the selected language (e.g., "at the vooruit" should display its English equivalent or the correct translation key should be applied).

## Actual Behavior
The tag remains static in Dutch ("at the vooruit"), regardless of the user's language preference.

## Acceptance Criteria
- [ ] The tag "at the vooruit" correctly translates based on the active locale.
- [ ] All related tags follow the same localization protocol and update correctly.
- [ ] Existing internationalization (i18n) tests pass without regression.

eliaryck:
## Problem
Tags and genres were always rendered using `display_name` (a static Dutch string), ignoring the active locale. Switching the UI language had no effect on chip labels. (at the vooruit for example)
 
### Bug fixes
| File | What changed |
|------|-------------|
| `ProductionCard.tsx` | `tags` prop changed from `string[]` to `Tag[]`; label now resolved via `getTranslatedRecord(tag.name, lang, fallback)` |
| `ProductionGridCard.tsx` | Genre chip label resolved with `getTranslatedRecord` + `labels` now passed as `genre.name` |
| `ProductionListCard.tsx` | Same fix as `ProductionGridCard` |
| `RelatedProductions.tsx` | Tag chip `name` prop switched from hardcoded `display_name` to resolved `tagName` |
| `SeriesDetailPage.tsx` | Removed inline `.map()` that extracted strings; passes raw `production.tags` + `lang` prop instead |
 
### Translation resolution order (all three components)
1. `name[lang]` — translated value for the active locale
2. `display_name` — fallback if translation key is missing
3. `String(id)` — last-resort numeric fallback
 
## Tests added
| File | New cases |
|------|-----------|
| `ProductionCard.test.tsx` | ~16 new cases covering: correct locale, `display_name` fallback, `id` fallback, null `name`, empty tags, duplicate labels, href targets, mixed translated/fallback tags |
| `ProductionGridCard.test.tsx` | 1 case: genre chip uses active locale (`en` -> "Dance", not "Dans") |
| `ProductionListCard.test.tsx` | 1 case: same as above for list card |


xxxxnoahboyxxxx (approved):

Floko2005 (approved):
The code looks good, all tests pass and when I check the components manually while running the frontend everything works.

Good job!

eliaryck:
@copilot resolve the merge conflicts in this pull request

Floko2005 (approved):
still works

PR #397, opened 2026-04-15
SeriesDetailPage doesn't use the Production Cards

### LINKED ISSUE #396: SeriesDetailPage doesn't use the Production Cards
The SeriesDetailPage currently doesn't use either ProductionListCard or ProductionGridCard and has its own component. It would be better to use the widely used components.

JDLMRN:
## Description
This PR lets the SeriesDetailPage use the ProductionListCard and ProductionGridCard components and removes it's local variant.

It also changes some minor styling to the page.

## Related Issues
- Closes #396

## Type of Change
- [ ] Feature
- [ ] Bugfix
- [x] Refactor
- [ ] Documentation

## Testing Instructions
Run all of the test and make sure they pass

Go to a serie (http://localhost:5173/series/95) to manually view and test the changes youself.

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests


xxxxnoahboyxxxx (changes requested):
- Are the Cards supposed to be slightly to the right?

Image

- You only show the genre chips? Where we not also going to show the other series tags a production might have (so not the current series tag).

JDLMRN:
> * Are the Cards supposed to be slightly to the right?

Yes, the alignment was intentional because of the timeline on the left

> * You only show the genre chips? Where we not also going to show the other series tags a production might have (so not the current series tag).

For me the chips are being loaded, could it be a problem with your database? 
image

xxxxnoahboyxxxx:
I mean the SeriesTag chips, not the genre chips.

JDLMRN:
> I mean the SeriesTag chips, not the genre chips.

Aha, my bad. In my latest commit this should be fixed.

xxxxnoahboyxxxx:
Yes, now the series tags are also there, but I meant only to show the other series tags, not the tag currently used for the series page you are in (also, maybe use the version that you can click so you open that series page (by making context series))

JDLMRN:
> Yes, now the series tags are also there, but I meant only to show the other series tags, not the tag currently used for the series page you are in (also, maybe use the version that you can click so you open that series page (by making context series))

I personally am not a fan of making the chips clickable in the cards itself. I would only make them clickable in the production details page and keep the rest of the styling the same across all of the other places where the Production cards are used. What's your thought?

xxxxnoahboyxxxx:
I just think it would be annoying to have to open the production first to be alle to click on the series tag, also then the chip would be Blue and you van distinguish between the genres and the tags

JDLMRN:
> I just think it would be annoying to have to open the production first to be alle to click on the series tag, also then the chip would be Blue and you van distinguish between the genres and the tags

I've made the chips clickable now, but I don't really like it when they are fully blue, it steals the attention. Is it fine to have it like this? 

The other option I was thinking of was to make the outline blue/purple to tell the difference.

xxxxnoahboyxxxx:
Changing the style like that also changes the style in other places where the chip is used.
How does the blue steal the attention? 
The genre tag is now also clickable (goes to homepage with that genre selected) (do we want this here). If the blue vs white is a problem, we can make it so the genre is purple.

prince-b-b:
I think the main issue is that the fully blue chips draw a bit too much attention compared to the rest of the card. Maybe we can make the visual treatment more subtle (e.g. outline or lighter background instead of full blue).

xxxxnoahboyxxxx:
If you all find it too eye catching, you can change it. But then find a way to do this without changing the styles/working in other places. Maybe use description for both and then update the styling for genre description to be purple too.

xxxxnoahboyxxxx:
@SELAB-2/viernulvier-team2

What do you guys think about how to show the difference between genre and tag chips in the seriesdetailpage?

JDLMRN:
I don't think that making the genre chips clickable too is a problem? I feel like it's either make them both clickable, or none.

And the changes I pushed indeed change the behavior of all ProductionCards, but I feel like it's good to have the cards all the same on every page. 

And for the coloring part, maybe this could work if you do want to differentiate?
image

eliaryck:
> I don't think that making the genre chips clickable too is a problem? I feel like it's either make them both clickable, or none.
> 
> And the changes I pushed indeed change the behavior of all ProductionCards, but I feel like it's good to have the cards all the same on every page.
> 
> And for the coloring part, maybe this could work if you do want to differentiate? image

Yes that's nice

xxxxnoahboyxxxx:
> I don't think that making the genre chips clickable too is a problem? I feel like it's either make them both clickable, or none.

That is fair.

> And the changes I pushed indeed change the behavior of all ProductionCards, but I feel like it's good to have the cards all the same on every page.

Yes, I agree that all the cards should look the same. I am talking about the production detail page and the search page.

On the search page, the chips get colored in when they are selected and should not be clickable (which they are now), so the edge being colored in might be confusing.

image

And on the production detail page, I find that the fully colored chips look better:

image


> And for the coloring part, maybe this could work if you do want to differentiate? image

Yes, this is fine, but personally, I just like the fully colored look better.

You are still showing the current tag. Is this not useless since you are already on that page? Just show all the other tags.

JDLMRN:
> On the search page, the chips get colored in when they are selected and should not be clickable (which they are now), so the edge being colored in might be confusing.

Why should they not be clickable there? If the tags are clickable too?

> And on the production detail page, I find that the fully colored chips look better:
> Yes, this is fine, but personally, I just like the fully colored look better.

I would keep the styling the same as much as possible, so I think it's fine to have the outline there too.
 
> You are still showing the current tag. Is this not useless since you are already on that page? Just show all the other tags.

I think the appearance of the Production cards shouldn't change depending on the page you are viewing them on. I think keeping them the same everywhere is better in my opinion.

xxxxnoahboyxxxx:
> > On the search page, the chips get colored in when they are selected and should not be clickable (which they are now), so the edge being colored in might be confusing.
> 
> Why should they not be clickable there? If the tags are clickable too?

Because they are also not supposed to be clickable there normally. On the search page they are just supposed to get colored when selected and show the colored ones first.

xxxxnoahboyxxxx (approved):

JDLMRN:
> Because they are also not supposed to be clickable there normally. On the search page they are just supposed to get colored when selected and show the colored ones first.

Yes, that's a good point, clicking them shouldn't route them to the same page with just that genre as a filter. I've changed that in my most recent commit.

prince-b-b (approved):

PR #398, opened 2026-04-16
Homepage

eliaryck:
I’ve created the homepage. Some tests have also been added; I suggest you just give it a go and try it out :)

I’ve also made some new changes, but that’s not really important. I’ve also fixed a bug where the current scroll position was retained, which meant that when you went to a new page, for example, you’d sometimes start at the bottom.

[Screencast from 2026-04-16 14-48-54.webm](https://github.com/user-attachments/assets/e01fb6c1-9599-47e5-92a3-c66d69744a55)


xxxxnoahboyxxxx (approved):
Looks good and all links work.

ArneMaegherman (approved):
Looks pretty on both desktop and mobile, all links works. Good job.

PR #413, opened 2026-04-21
Change function to work with changing query params also

### LINKED ISSUE #412: [BUG] When using pagination, you do not go to the top of the page.
## Problem Statement
With pagination, you don’t go to the top of the page. There’s a function for that, but a small adjustment is needed so that it still works when the query parameters change.

eliaryck:
With pagination, you don’t go to the top of the page. There’s a function for that, but a small adjustment is needed so that it still works when the query parameters change. I think it’s my smallest PR ever ;)


xxxxnoahboyxxxx (approved):
That must have taken a lot of work. ;D

ArneMaegherman (approved):
Nicely done

PR #417, opened 2026-04-22
Add max warnings 0

### LINKED ISSUE #416: [BUG] ESLint warnings to error
## Problem Statement
At the moment, the ESLint warnings are not treated as errors, and it would be better to do so.

eliaryck:
I think it should now check the format and linting for every PR. And the workflow works, because I’ve had errors with it myself. :)


tobitdsm (approved):
Alright, looks good!

JDLMRN (approved):

PR #451, opened 2026-04-27
[BUG] Not all genres and tags are displayed in the filters.

### LINKED ISSUE #425: [BUG] Not all genres and tags are displayed in the filters.
## Problem Statement
Not all genres and tags are displayed in the filters. This is because the services only return the first page. 

## Scope
Frontend

## Expected Behavior
All tags/genres need to be showed.

## Actual Behavior
Not all are shown.

## Acceptance Criteria

JDLMRN:
## Description
Adds pagination support to the genre fetching on the ProductionsPage.

## Related Issues
- Closes #425

## Type of Change
- [ ] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation

## Testing Instructions

Open /archive and check if all genres are present in the filter panel

Make sure all tests still pass

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests


ArneMaegherman (commented):
The tags and genres list are now complete. The tests and linting succeeds.

Maybe change the order of the filter panel options, the genres tab is now very long while the attendance and performer type will always be small.

Also I like the change regarding the centering but on some screen sizes is still doesn't look that great.

Image

JDLMRN:
> Maybe change the order of the filter panel options, the genres tab is now very long while the attendance and performer type will always be small.

I put those two options intentionally last and collapsed them by default because I don't really see their purpose. Do you think they are worth putting higher than the genres and tags?

> Also I like the change regarding the centering but on some screen sizes is still doesn't look that great.

I personally think it looks fine this way. We came to the conclusion that centering them there looks weird and I feel like shrinking the searchbar and sorting controls when there is only one row would just waste the space on the sides. What would your proposal be on this width?

frontend/src/components/CollectionPageLayout.tsx
```diff
@@ -302,18 +302,11 @@ const CollectionPageLayout = ({
   return (
     <Box sx={{ py: { xs: 3, md: 4 } }}>
       <Container maxWidth="xl">
-        {/*
-          Single layout for every page: the (optional) sidebar and the
-          content column live in one flex row. When the sidebar is visible
-          we left-align the block so the filter panel hugs the container's
-          left edge; otherwise the content column (and any free space from
-          its stepped cap) center horizontally in the container.
-        */}
```
xxxxnoahboyxxxx, line 311: Did you mean to remove this comment?

xxxxnoahboyxxxx (commented):
I do indeed see all genres now.
It might be good to try seeing if it is better to allow multiple genres per line because now the list is very long and there is a lot of whitespace. (try without the cross, so the chip does not become bigger). Or maybe something else to impove the list of chips ( a search field, collapsable subsections (alfabet), ...).

Image

frontend/src/components/CollectionPageLayout.tsx
JDLMRN: Yes, because it doesn't apply anymore after changing:

---

`justifyContent: sidebarInline ? 'flex-start' : 'center',`

to 

`justifyContent: 'center',`

---

Now the full container is always centered, not depending wether or not there is a sidebar.

ArneMaegherman:
> > Maybe change the order of the filter panel options, the genres tab is now very long while the attendance and performer type will always be small.
> 
> I put those two options intentionally last and collapsed them by default because I don't really see their purpose. Do you think they are worth putting higher than the genres and tags?

I just thought that they would be more visible when put higher, but if it already is a deliberate choice you can keep it like this.
 
> > Also I like the change regarding the centering but on some screen sizes is still doesn't look that great.
> 
> I personally think it looks fine this way. We came to the conclusion that centering them there looks weird and I feel like shrinking the searchbar and sorting controls when there is only one row would just waste the space on the sides. What would your proposal be on this width?

Once more, if this is a deliberate choice it is fine. I just thought it looks kinda weird but I don't really see a easy fix.

ArneMaegherman (approved):
Looks good, well done.

xxxxnoahboyxxxx (approved):

Issue #438, opened 2026-04-23
[BUG] Tag chips do not get colored


xxxxnoahboyxxxx:
## Problem Statement
The tag chips in the cards do not get colored in when a tag is selected in the filter panel.

## Scope
archive page

## Steps to Reproduce
1. Select a tag in the filter panel 

## Expected Behavior
The chips should get colored in blue

## Actual Behavior
Nothing happens

## Screenshots

<img width="1492" height="858" alt="Image" src="https://github.com/user-attachments/assets/607ecb55-8f73-4528-8272-508cb61da25e" />

## Acceptance Criteria
- [ ] Bug is fixed and verified
- [ ] Existing tests still pass

Issue #437, opened 2026-04-23
[FE][FEATURE] Use URL for language


xxxxnoahboyxxxx:
## Problem Statement
Make language work via the UTL instead of local storage.

## Scope
Navbar
URL state

PR #455, opened 2026-04-28
[BUG] only 1 tag/genre used when filtering

### LINKED ISSUE #439: [BUG] only 1 tag/genre used when filtering
## Problem Statement
Only 1 tag/genre gets used when filtering the productions, even when multiple are selected.

## Scope
The archive page

## Steps to Reproduce
1. Select muliple genres or tags in the filter panel
2. check if results found changes

## Expected Behavior
The cards should get updated.

## Screenshots

Image

--- 

Image

## Acceptance Criteria

JDLMRN:
## Description
Makes sure that when you select two or more genres/tags, it actually filters on all of them instead of just the first selected one.

## Related Issues
- Closes #439

## Type of Change
- [ ] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation

## Testing Instructions
- Make sure all automatic tests pass
- Go to the productions page 
- Filter on multiple genres/tags

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests


xxxxnoahboyxxxx:
The coloring bus is already fixed in another branch, because that bug has already been addressed.
issue #438 
PR #437

xxxxnoahboyxxxx (approved):
Filtering on multiple tags/genres works. Good job.

JDLMRN:
> The coloring bus is already fixed in another branch, because that bug has already been addressed. issue #438 PR #437

Whoops, didn't know. Reverted those changes so they don't interfere with your fix.

xxxxnoahboyxxxx:
The code looked almost exactly the same so i don't think it would have caused problems

ArneMaegherman:
The filtering works like we want, nice job.  

In the screenshot I filtered on 'Nightlife' and 'Te gast'. Like you can see, the filtering works but the chips which are colored are not given priority in displaying them on the page. Would it be possible to fix this or would this be a hard problem?

image
image

JDLMRN:
> In the screenshot I filtered on 'Nightlife' and 'Te gast'. Like you can see, the filtering works but the chips which are colored are not given priority in displaying them on the page. Would it be possible to fix this or would this be a hard problem?

I wouldn't display them in the production details itself (your last screenshot), since you're not on the page with the filter panel anymore. However, putting them at the front in the cards is a good idea! I've implemented that in my last commit.

ArneMaegherman (approved):
Looks good, well done.

> I wouldn't display them in the production details itself (your last screenshot), since you're not on the page with the filter panel anymore. However, putting them at the front in the cards is a good idea! I've implemented that in my last commit.

You did exactly what I was trying to suggest. The last screenshot was just to showcase that the production also had both tags, not to color them in the detailpage.

PR #468, opened 2026-05-04
[BE] Correct upload experience for media files

### LINKED ISSUE #467: [BE] [BUG] Correct upload experience for media files
## Problem Statement
The CMS file upload experience for media files has poor UX and inconsistent validation behavior. 

Currently:
- The file picker allows selecting unsupported image formats (e.g. GIF, BMP), even though the backend only accepts JPEG, PNG, and WebP.
- When a user selects an unsupported file, an error is shown but does not clear when a new file is selected, leading to confusing feedback.

This results in unnecessary validation errors and a frustrating upload experience for administrators.

## Scope
- Restricting the file picker to only allow supported image formats (JPEG, PNG, WebP)
- Ensuring validation errors are cleared when a new file is selected
- Improving UX feedback during file selection
- Aligning frontend validation with backend-supported file types

## Acceptance Criteria
- [ ] File picker only allows selecting supported formats (JPEG, PNG, WebP)
- [ ] User cannot select unsupported file types via the file browser UI
- [ ] Validation error is shown when an unsupported file is selected
- [ ] Validation error is cleared immediately when a new file is selected
- [ ] Valid file selection does not retain previous error state
- [ ] Upload flow behaves consistently and predictably

prince-b-b:
## Description
Improves the UX of media file uploads in the Django admin (CMS).

Changes include:
- Restricting the file picker to only allow supported file types (JPEG, PNG, WebP, PDF)
- Adding client-side validation for drag & drop uploads (since `accept` does not cover this)
- Displaying validation errors in Django admin style (instead of browser alerts)
- Automatically clearing validation errors when a new file is selected

This ensures that users receive immediate and consistent feedback when uploading invalid files, reducing frustration and avoiding unnecessary backend validation errors.

## Related Issues
- #467 

## Type of Change
- [x] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation

## Testing Instructions
Start the Django backend and do some testing in the CMS.
   ```bash
   python manage.py runserver


Floko2005 (approved):
Nice, everything works like expected. I tested each of the allowed file types and they are all accepted. The drag and drop also works, and when I dropped a file with the wrong extension I received an error. The code also looks good.

Good job!

tobitdsm (changes requested):
This is indeed very good, but the image upload in the tag edit page doesn't follow the specifications you provided. Either this needs to be changed to work like the media upload, or we use the media-item/gallery association like how the productions work.

Floko2005:
I do agree with @tobitdsm that it would be nice if this can be expanded to all pages with files in the cms. This was not the scope of this PR , but you can do it here or maybe open a different PR after merging this one.

prince-b-b:
> ... the image upload in the tag edit page doesn't follow the specifications you provided. Either this needs to be changed to work like the media upload, or we use the media-item/gallery association like how the productions work.

I’ve just implemented this using the existing media gallery pattern (like productions) to keep things consistent and avoid a separate image upload flow. I’d appreciate you guys' feedback to see if this matches your expectations.

Floko2005 (changes requested):
I don't think using the existing media_gallery pattern here is the way to go. The media_gallery patten is used to link multiple media items to the tag.

A tag can only show one image in the frontend (it's shown in the card) so using this pattern is really not necessary. I think what Tobit meant is to keep the singe file upload, but using your media_file_upload.js for the drag and drop and limiting the valid file extensions for the upload (for tags pdf should also not be allowed).

The way it worked before is that if there is an uploaded image, the tag shows that image in the frontend. Otherwise it uses an image from the most recent production that contains that tag. It should still work like this.

backend/apps/tags/admin.py
```diff
@@ -44,27 +60,11 @@ def get_queryset(self, request: HttpRequest) -> QuerySet:
...
-    identify which tags are active and what category they belong to.
-
-    ``search_fields`` includes ``type`` and ``source`` to allow autocomplete
-    from :class:`~apps.productions.admin.ProductionTagInline`.
-
-    Queryset strategy
-    -----------------
-    ``prefetch_related("translations")`` prevents N+1 queries when the inline
-    renders all language translations on the detail page.
-    """
```
Floko2005, line 67: I don't think these comments should be removed

backend/static/admin/js/media_file_upload.js
Floko2005, line 1: Maybe add some extra docs already so we don't need to do this later

Floko2005 (changes requested):
I also forgot about the blogs! You should add this to the image field for blogs too. I also left some other remarks.

tobitdsm (approved):
This looks good now, I think this is everything. Well done!

Floko2005 (approved):
Nice! Works like it should and the code looks good now. Good job

PR #469, opened 2026-05-05
444 [FE][BUG] add chips and timestamp to productioncards in relatedproductions

### LINKED ISSUE #444: [FE][BUG] Add chips and timestamp to ProductionCards in RelatedProductions
## Problem Statement
Currently the timestamp and tags chips don't get shown in the ProductionCards in the RelatedProductions component.

## Scope
This is visible on pages containing the RelationsProductions component. These are BlogDetailPage and ProductionDetailPage.

## Steps to Reproduce
1. Open a detailpage containing the RelationProduction component.

## Expected Behavior
We want the timestamp and chips to be displayed in the card.

## Actual Behavior
Currently only the title and artist name.

## Acceptance Criteria
- [ ] Regression test added (Jest) OR existing test updated
- [ ] No new console errors introduced

ArneMaegherman:
# Add chips and timestamps to productioncards in relatedproductions

## Description
This PR fixes the bug where the time data and tags/genre chips didn't get shown on the productiongridcards when these were displaying the relatedProduction on the BlogDetailPage or ProductionDetailPage.

## Related Issues
- Closes #444 

## Type of Change
- [x] Feature
- [x] Bugfix
- [ ] Refactor
- [ ] Documentation

## Testing Instructions
- In admin panel: 
    - Make sure your local database contains a tag
    - Tag a certain amount of productions with this tag
    - Create a blog post and add these same productions to the Linked Productions field
- On site:
    - Take a look at the detailpage of the blog and productions you made/updated.
-  Run tests and linting on both frontend/backend.

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests

## Note
I (Arne) did most of the work on this PR but I added Prince-Bert as second assignee because he probably is more in tune with the backend changes in this PR.


backend/apps/productions/serializers.py
Floko2005, line 1: I found it a bit weird that you were using the through table (ProductionTag) that defines the m2m relationship between Productions and Tags, but then I found an issue with the code that probably caused some extra confusion. Someone added a ProductionTagTranslation table which is never used so it should be removed (in a different PR). So after seeing that I understand why you would use this through table.

You should change the code so ProductionTag.translation is never used, because it will be removed. To get the translations of a tag, you should use the Tag model in ```apps.tags.models```. I think ProductionTag can be completely replaced by just using the Tag model. The code will probably be simpler as well in that case.

If you use the Tag model getting the related productions would look something like this:

```python
def _get_related_productions_qs(self, tag: Tag, exclude_id: int):
        """Return a queryset of productions sharing the given tag, excluding the current production."""
        return (
            Production.objects.filter(tags=tag)
            .exclude(id=exclude_id)
            .select_related("uit_database_type", "media_gallery")
            .prefetch_related(
                "translations__language",
                Prefetch(
                    "productiongenre_set",
                    queryset=ProductionGenre.objects.select_related("genre").order_by("position"),
                    to_attr="prefetched_production_genres",
                ),
                Prefetch(
                    "tags",
                    queryset=Tag.objects.prefetch_related("translations__language").order_by("type", "id"),
                    to_attr="prefetched_tags",
                ),
                Prefetch(
                    "media_gallery__media_items",
                    queryset=MediaItem.objects.prefetch_related("translations__language", "crops").order_by("position"),
                ),
            )
            .annotate(
                first_event_start=Min("events__starts_at"),
                last_event_end=Max("events__ends_at"),
            )
        )

    def get_related(self, obj: Production) -> list | None:
        """Return related productions grouped by the tags on this production."""
        if "related" not in self.context.get("include", set()):
            return None

        tags = getattr(obj, "prefetched_tags", None) or list(
            obj.tags.prefetch_related("translations__language").order_by("type", "id")
        )
        if not tags:
            return []

        return [
            {
                "tag": RelatedTagSerializer(tag, context=self.context).data,
                "productions": RelatedProductionSerializer(
                    self._get_related_productions_qs(tag, obj.id),
                    many=True,
                    context=self.context,
                ).data,
            }
            for tag in tags
        ]
```
**Note:** I'm not sure if this works 100% correctly, you should test it first (but at first sight it seems to work).

You probably need to edit the tests you wrote for this, I did not look at those.

Floko2005 (changes requested):

frontend/src/types/Productions.ts
JDLMRN, line 1: I would let the Production inherit from the RelatedProduction interface. That way there is less duplication of fields and it's better maintainable.

JDLMRN (changes requested):
I'm not an expert in the backend side of things so I'm gonna let @Floko2005 handle that, but I did have one tiny suggestion I left as a comment in the frontend/src/types/Productions.ts file.

Apart from that, everything seems to be working, well done!

Floko2005:
@ArneMaegherman I fixed the merge conflicts for you. I tested it and it doesn't seem that it broke anything but you should also check it to be sure.

I did not check the frontend and backend linting stuff, you will have to fix those.

JDLMRN (changes requested):
Thanks for letting Production inherit from the RelatedProduction interface.

I'm pretty sure this worked the last time I reviewed this PR, but the related productions on the ProductionDetailPage don't show any chips or timestamps. (e.g. `/nl/producties/2374`)

Image

If you can address that along with the linting error, it's good to go

ArneMaegherman:
> Thanks for letting Production inherit from the RelatedProduction interface.
> 
> I'm pretty sure this worked the last time I reviewed this PR, but the related productions on the ProductionDetailPage don't show any chips or timestamps. (e.g. `/nl/producties/2374`)

> If you can address that along with the linting error, it's good to go

Did you run the backend locally? Because there are needed changes that the dev backend does not have at the moment.

ArneMaegherman:
> @ArneMaegherman I fixed the merge conflicts for you. I tested it and it doesn't seem that it broke anything but you should also check it to be sure.
> 
> I did not check the frontend and backend linting stuff, you will have to fix those.

Thank you for the merge. I've fixed the linting and for me everything works.

backend/tests/productions/test_production_serializers.py
Floko2005: This can be removed.
```suggestion

```

Floko2005 (changes requested):
It works for me! This is shown on the blog page for a related production:

Image

This is shown on the production detail page:

Image

There are just some things that can be removed in the test_production_serializers.py, I added them as suggestions.

JDLMRN (approved):
> Did you run the backend locally? Because there are needed changes that the dev backend does not have at the moment.

This is completely my fault, I ran it with the production as backend without realizing that there were made changes to it.
Everything looks good to me, well done!

Floko2005 (approved):
Good job!

PR #473, opened 2026-05-06
Show only prods/events before today

### LINKED ISSUE #448: [LAYER][FEATURE] Fetch data before current date
## Problem Statement
At present, we also retrieve productions for the archive after a certain date. We had to change this to include only those that have passed that date.

## Scope
Backend

## Design/Technical Approach
It's a small .filter() that needs to be added.

eliaryck:
I have ensured that the archive page only displays productions and events on the details page that are from before the current time and date. The date ranges have also been adjusted. The same applies to series. Future events are now also not displayed on production detail pages. I think I need to update some code when PR #469 is merged. 
For example, **previously** on the server: 
image
image

And **now**:
image
image

I have also set the coverage back to 100%.

### DIFF
--- a/backend/apps/productions/filters.py
+++ b/backend/apps/productions/filters.py
@@ -1,6 +1,7 @@
 """Filters for the Productions app."""
 
 import django_filters
+from django.db.models.functions import Now
 
 from apps.core.filters import BaseModelFilter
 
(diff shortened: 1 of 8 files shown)


prince-b-b (changes requested):
Great job, Elias! The feature works as it's supposed to on my end. I did notice something regarding the event filtering: this happens once in the prefetch in `retrieve()` and again in `get_events()`. Couldn't this cause extra queries? If so, I'd use something in the sense of `Prefetch(..., to_attr="prefetched_past_events")` and read that list in the serializer instead.

(The files that contain the code in question are `productions/views.py` and `productions/serializers.py`.)

eliaryck:
> Great job, Elias! The feature works as it's supposed to on my end. I did notice something regarding the event filtering: this happens once in the prefetch in `retrieve()` and again in `get_events()`. Couldn't this cause extra queries? If so, I'd use something in the sense of `Prefetch(..., to_attr="prefetched_past_events")` and read that list in the serializer instead.
> 
> (The files that contain the code in question are `productions/views.py` and `productions/serializers.py`.)

Good point – that could indeed cause problems

prince-b-b (approved):

Floko2005 (approved):
The code looks good and everything works like expected.
I tested it by changing the events of a production to be in the future, and as expected the production was removed from the archive, seriesdetails and blogdetails page.

Good job

PR #475, opened 2026-05-06
Add include for blogs in backend and frontend

### LINKED ISSUE #474: Add included for related productions
## Problem Statement  
Currently, when visiting pages such as blogs, related products are fetched through separate API requests. This results in multiple network calls for a single page load, increasing latency and reducing performance.  

We want to optimize this by allowing related products to be included directly in the initial request via a query parameter, so that all required data is fetched in a single request.

---

## Scope  
**Included:**  
- Adding support for an `include` (or similar) query parameter in API requests (e.g. `?include=related`)  
- Updating backend endpoints to optionally embed related products in the response  
- Updating frontend to use the new parameter and remove redundant fetch calls  

**Excluded:**  
- Major restructuring of the API  
- Changes to unrelated endpoints  
- Caching strategies (handled separately)

---

## Design / Technical Approach  
Extend existing endpoints to support an optional query parameter (e.g. `include=related`).  

When present:  
- The backend resolves and embeds related product data in the main response  

When absent:  
- The endpoint behaves as before (no breaking changes)  

Frontend will be updated to request related products via this parameter instead of making a separate call.

eliaryck:
I’ve added that `included` so that you only send one request, rather than separate ones for each blog. See the issue description. The issue with the one missing line for 100% coverage is being resolved in another pull request (#473 ) of mine.

You can check this by going to `Inspect` on the page and then selecting `Network`. (on details of a production)

### DIFF
--- a/backend/apps/productions/schemas.py
+++ b/backend/apps/productions/schemas.py
@@ -289,8 +289,9 @@ _PRODUCTION_RETRIEVE = extend_schema(
         "per language; the dictionary is empty when no descriptions have been added. "
         "`include=related` returns productions grouped by tag, where each related tag "
         "is compact and only exposes `id`, `name`, and `display_name`. "
-        "Use `?include=events` to include all related events in the response, or "
-        "`?include=related` to include related productions grouped by tag."
+        "Use `?include=events` to include all related events in the response, "
+        "`?include=related` to include related productions grouped by tag, and "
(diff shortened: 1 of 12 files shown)


xxxxnoahboyxxxx (approved):
Looks good

Floko2005 (approved):
Works like expected when I test it. I don't see any big issues with the implementation.
Good job!

PR #510, opened 2026-05-14
Add tests and delete old file and transfer images to webp

### LINKED ISSUE #509: [BUG] old crops are never deleted
## Problem Statement
Old crop image files are never deleted from storage when a crop is re-fetched and overwritten. Every sync leaves orphaned files behind.

## Scope
- Storage (S3 / local file system)
- `_save_single_crop` in `viernulvier_media_sync.py`

## Steps to Reproduce
1. Run crop sync -> crop image saved at `path/a.jpg`
2. Run crop sync again -> new image saved at `path/b.jpg`, DB updated
3. `path/a.jpg` remains in storage, unreferenced

## Expected Behavior
The old file is deleted from storage once the DB record points to the new one.

## Actual Behavior
Old files accumulate in storage indefinitely.

## Acceptance Criteria
- [ ] Old file is deleted after a successful `update_or_create`
- [ ] New file is cleaned up if the DB write fails (no orphan on error either)

## Discussion
Fix: fetch `existing_image_path` before saving, delete it inside the transaction
after the DB update succeeds. Wrap the DB write in a try/except that deletes
the newly saved file if the transaction fails.
Requires `import contextlib`.

eliaryck:
## Overview

Fixes orphaned crop image files accumulating in storage on every sync, and converts downloaded crop images to WebP where it saves space.

## WebP Optimization Results

The implementation of WebP conversion provides significant storage savings:

* **Total Storage Reduction:** ~65.3% average saving across synced crops.
* **JPEGs:** Average **41%** smaller with no perceptible loss in quality.
* **PNGs:** Massive gains of approximately **93%** (e.g., reducing 670KB to 45KB).
* **Impact:** Average file size decreased from ~295KB to ~102KB per item, significantly improving load times and reducing bandwidth costs.

## Changes

**`viernulvier_media.py`**
- Added `_maybe_convert_to_webp` - converts crop images to lossy WebP (q82); skips animated GIFs, already-WebP URLs, and cases where WebP would be larger
- Updated `_save_single_crop` to fetch the existing file path before overwriting, delete it after a successful DB update, and clean up the newly saved file if the DB write fails

**`test_viernulvier_media.py`**
- Added 7 unit tests covering `_maybe_convert_to_webp`: short-circuit on `.webp` URL, animated skip, size guard, successful conversion, alpha/palette mode normalisation, and unexpected exceptions

## Why

Previously every sync saved a new file without removing the old one, causing unbounded storage growth. WebP conversion is a bonus win on file size.

### DIFF
--- a/backend/apps/imports/scrapers/viernulvier_media.py
+++ b/backend/apps/imports/scrapers/viernulvier_media.py
@@ -2,12 +2,17 @@
 
 from __future__ import annotations
 
+import contextlib
+from io import BytesIO
+
 import logging
+import os
 import re
 import time
 from typing import TYPE_CHECKING, Any, cast
 from urllib.parse import urljoin
 
+from PIL import Image, UnidentifiedImageError
 from django.core.exceptions import FieldDoesNotExist
 from django.core.files.base import ContentFile
 from django.db import transaction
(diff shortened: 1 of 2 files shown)


prince-b-b (approved):
Nicely done, Elias!

tobitdsm (approved):
Very nice, good catch that we previously didn't delete files that should have been overwritten. This will also improve load times!

PR #511, opened 2026-05-14
497 bug cannot right click a production card

### LINKED ISSUE #497: [BUG] Cannot right-click a production card
## Problem Statement
You can not right-click on a production card

## Scope
productionpage

## Steps to Reproduce
1. Go to the Archive tab
2. right-click a card
3. see no: `open in new tab`

## Screenshots

Image

## Acceptance Criteria

ArneMaegherman:
## Description
Fix production card chip interactions on the archive/production page so cards remain right-clickable and support browser link actions. The production cards keep their outer `RouterLink`, while chips inside cards stay interactive without creating nested anchor violations. This also preserves chip navigation behavior on detail pages and updates genre chip clicks in cards to append to existing genre filters instead of replacing them.

## Related Issues
- #497

## Type of Change
- [x] Bugfix

## Testing Instructions
Run the focused frontend regression suite from frontend:

```bash
npm test -- --runInBand src/__tests__/components/GenreAndTagChip.test.tsx src/__tests__/components/ProductionGridCard.test.tsx src/__tests__/components/ProductionListCard.test.tsx src/__tests__/pages/SeriesDetailPage.test.tsx
```

Manually and visually check the chips display and behavior in the filterpanel, archive page, seriesdetailpage and related productions.  

This verifies that production cards no longer block right-click/open-in-new-tab behavior, chips remain interactive inside cards, and genre filter clicks append to existing filters.

## Checklist for Reviewers
- [ ] Follows Style Guide
- [ ] Added Unit tests

## Note:
I tried solving the issue while still keeping the chips clickable in the ProductionPage if this is not what we want I can remove this functionality.


xxxxnoahboyxxxx (commented):
- You can indeed right-click the card now. You can also right-click the series chip, but only in grid mode. In list view, the series chip is not right-clickable.

- The tests also currently fail.

xxxxnoahboyxxxx (commented):
All right-clicks now go to the detail pages. 

Did you make it so that clicking a genre in a card selects it in the filter? But cannot deselect it?
Because I would either do
- Series is a link, and genre does nothing
or 
- You can select genres and series in the cards to add or deselect them from the filters

So it is not ambiguous that genre does add to the filter, but series sends you to the detail page. What do you think?

Also, don't forget to look at the copilot feedback.

frontend/src/components/chips/GenreAndTagChip.tsx
```diff
@@ -130,7 +178,11 @@ const GenreAndTagChip = ({
       aria-pressed={context === 'search' ? selected : undefined}
       aria-label={resolvedAriaLabel}
       tabIndex={isClickable ? 0 : -1}
-      {...(linkTo ? { component: RouterLink, to: linkTo } : { component: 'div' })}
+      {...(linkTo
+        ? disableLink
+          ? { component: 'div', onClick: handleLinkClick }
```
JDLMRN:
```suggestion
          ? { component: 'div' }
```

Typescript gave me the follwing error:
`'onClick' is specified more than once, so this usage will be overwritten.`

Removing this should fix this without changing the behaviour as far as I can see.

JDLMRN (changes requested):
The problem is indeed fixed, good job.

My IDE did give me some error which I added a fix for as a suggestion.

xxxxnoahboyxxxx (commented):
It is now consistent everywhere indeed. 
Is it on purpose that you can only select a chip in the cards but not deselect them?
Don't forget to take a look at the copilot feedback.

JDLMRN:
> Is it on purpose that you can only select a chip in the cards but not deselect them?

The thing is that if you make them deselectable too, then you need to use the style with the cross, which makes the layout of the chips shift inside of the cards.

I think before this they were just not selectable/clickable at all.

xxxxnoahboyxxxx (approved):

JDLMRN (approved):

PR #537, opened 2026-05-17
Fix productions view with no events and correct tests

### LINKED ISSUE #536: [BUG] Productions without events
## Problem Statement
Productions without events were included in query results and landing stats, which is incorrect for an archive.

## Changes
- Removed `~Exists` fallback from the production queryset filter and `landing_stats` count
- Productions without any events are now excluded everywhere

eliaryck:
The data now looks **much better**, and there is more of it than before because events with just a start time are now also displayed. (CSV data)

This PR fixes a bug where productions without any associated events were failing to display or break logic in the backend views and homepage counters. The query filters have been refactored to cleanly handle both past events and open-ended or event-less productions using reusable Q expressions. Additionally, the test suite has been updated to align with these query constraints. 

Test cov is still 100% :)


tobitdsm (approved):
This looks good, and is a nice improvement. Well done!

prince-b-b (approved):

PR #540, opened 2026-05-17
Make footer same color as navbar

### LINKED ISSUE #539: Footer same color as navbar
## Problem Statement
The navigation bar currently shares the same background color as the rest of the page, making it blend in completely. This reduces visual hierarchy and makes it difficult for users to distinguish the navigation area from the content.

## Scope
- **Included:** Update the navbar background color to a solid black.
- **Excluded:** Changing the layout, navigation links, or styling of other global components.

eliaryck:
Updated the navbar background to solid black using theme tokens for better contrast and separation from the page content.

Before: 
image

After: 
image


ArneMaegherman (approved):
Looks good

Floko2005 (approved):
Looks much better, now there is a difference between the darkmode background and footer.

PR #559, opened 2026-05-19
show count bug fix

Floko2005:
This PR fixes a bug that the show count button in the cms filter panel leads to an error 500. This bug only occurred on the production page because it has a custom filter panel.

This bug should now be fixed, test it by opening the production list page and press the show count button. You should see the count appear for the attendance mode and performer type filters.


tobitdsm (approved):
This looks good, simple but effective fix. Well done!

eliaryck (approved):
Hopefully last bug :)

PR #560, opened 2026-05-19
fixed the 'total selected' value issue in cms

Floko2005:
This PR fixes the issue that the total selected does not update when using the 'select all' in the cms:

[Screencast from 2026-05-19 15-57-08.webm](https://github.com/user-attachments/assets/06ca90b4-12a4-47ee-ac7c-822f14a1359d)

### DIFF
--- a/backend/apps/core/admin.py
+++ b/backend/apps/core/admin.py
@@ -122,11 +122,18 @@ class PersistentSelectionMixin:
         is_select_across = False
 
         if request.method == "POST":
-            merged_ids = self._inject_persisted_selection_into_post(request)
             is_select_across = request.POST.get("select_across") == "1"
 
-            if merged_ids and not is_select_across:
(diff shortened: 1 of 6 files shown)


eliaryck (changes requested):
If you select everything and then add a tag, for example, that will only apply to the items on the current page.

Floko2005:
> If you select everything and then add a tag, for example, that will only apply to the items on the current page.

This should be fixed now, but when you try to add it to a lot of tags you get limited by django so the select all to add tags or genres isn't that useful...

It does work for deletion, but it takes some time to fetch everything. It just limits the amount a user can post.

eliaryck (approved):
I didn't spot any issues. Cov is still 100% :)

tobitdsm (approved):
Looks good, hopefully this was the last one!

